PR #3, opened 2020-06-03
update ld-sw framework according to new boot implementation; cleanup bootloader Makefile; update AES-KU040-DB-G FPGA synthesis framework; cleanup firmware Makefile

JDLopes:


jjts:
Nice job

PR #8, opened 2020-06-09
improve and fix FPGA synthesis framework; remove FPGA synthesis warnings; update interconnect submodule

JDLopes:


rtl/src/int_mem.v
```diff
@@ -76,7 +76,7 @@ module int_mem
         //cpu slave interface
         //no address bus since single address
         .cpu_valid(boot_ctr_req[`valid(0)]),
-        .cpu_wdata(boot_ctr_req[`wdata(0)]),
+        .cpu_wdata(boot_ctr_req[`wdata(0)-(`DATA_W-2)]),
```
jjts, line 79: need to change to `wdata(0,2): macro should accept data width

PR #9, opened 2020-06-17
add darkrv submodule; add darkrv CPU to Makefiles

JDLopes:


jjts:
Darkrv is private and not visible from iob-soc. This was discussed in the chat window. I will create a special soc to test darkrv. Until then use your repo.

It was also discussed that the source files of submodules are no longer listed in SoC. Create src.mk in darkrv to be included in iob-soc-darkrv

PR #16, opened 2020-07-03
TIMER integration step

ChrisEdgley:
Initiating integration of TIMER, modifications to some SoC files, tex md corrections, peripherals MSB bug fixed


jjts:
This repo is not to contain a timer ....

PR #27, opened 2020-07-14
asic memory restructured

josejsarmento:
- BOOTRAM is now SRAM
- deleted .sh scripts


hardware/asic/umc130/memory/Makefile
```diff
@@ -1,17 +1,18 @@
-FIRM_DIR:=../../../../software/firmware
-BOOT_DIR:=../../../../software/bootloader
+ROOT_DIR:=../../../..
+include $(ROOT_DIR)/system.mk
 
-BAUD:=115200
-FREQ:=100000000
+# in the future, BAUD and FREQ is to be defined by system.mk; this .mk file still needs to be changed for this end
+BAUD_TMP:=115200
```
jjts, line 5: future is now, this is already defined, use it

jjts: BAUD and FREQ ar probably not used in the hardware. Try deleting them. As far as I remember only software and testbench needs these vars that are used to init the UART in run time.

hardware/asic/umc130/memory/Makefile
```diff
@@ -1,17 +1,18 @@
...
+FREQ_TMP:=100000000
 
-all:
-	make -C $(FIRM_DIR) BAUD=$(BAUD) FREQ=$(FREQ)
-	make -C $(BOOT_DIR) BAUD=$(BAUD) FREQ=$(FREQ)
+run:
+	make -C $(FIRM_DIR) BAUD=$(BAUD_TMP) FREQ=$(FREQ_TMP)
```
jjts, line 9: No need to build the firmware or bootloader here. This is done at the top-level makefile. Here we just make mem.

jjts: get inspeired in fpga build

hardware/asic/umc130/memory/Makefile
josejsarmento: BAUD and FREQ deleted

hardware/asic/umc130/memory/Makefile
josejsarmento: firmware and bootloader have no reason to be here, deleted

PR #81, opened 2020-12-12
add ASIC flow; fix synthesis error in boot_ctr module; fix memwapper_make script

JDLopes:


jjts:
I think asic simulation must go into hardware/simulation, but I'm merging anyway as this is a good addition

PR #86, opened 2020-12-16
PC emulation now available for hello world

AndreMerendeira:


Makefile
```diff
@@ -263,7 +274,8 @@ endif
 # CLEAN ALL
 clean-all: sim-clean fpga-clean asic-clean board-clean doc-clean
```
jjts, line 275: add pc-clean here

AndreMerendeira: done

PR #90, opened 2020-12-31
console goes into a loop to recv requests after loading the firmware,…

AndreMerendeira:
… exits when receives EOT


AndreMerendeira:
UART must be updated after merge

jjts:
PR to branch console_proxy instead 

make sure make sim-test and make fpga-test work before PRing

PR #97, opened 2021-01-26
Multi inst

AndreMerendeira:


jjts:
soc is agnostic about peripherals , cannot contain uart related stuff

PR #103, opened 2021-02-05
Verilator testbench seems to be working

slahiruk:
Testbench with clk and reset signal seems to be working.


jjts:
The idea is to make a long loop that ticks the clock for a long while until we see the UART sending the ENQ byte repeatedly

PR #104, opened 2021-02-06
improve software.mk for selecting march compilation flags

JDLopes:


jjts:
what is the point of this commit?

JDLopes:
> what is the point of this commit?

Just to get a clean solution for selecting march compilation options.

jjts:
seems less readable and PR to master branch are now totally disabled

PR #108, opened 2021-02-18
Verilator updates to wrapper

slahiruk:
Just putting in this PR so that if things are not as desired you can comment and have a look at what's going on


jjts:
This wrapper is empty! If this is the system top, where is system within it? Did you look at the fpga wrapper?

PR #109, opened 2021-02-18
Verilator

slahiruk:
VCD file is generated but not running like the previous testbench, seems to have halted


jjts:
closing this PR and expecting a more verilator specific one

PR #110, opened 2021-02-18
Verilator changes to include wrapper

slahiruk:
The changes included in enabling the wrapper are not included in hardware.mk


hardware/src/sim_system_top.v
```diff
@@ -25,18 +25,16 @@ module sim_system_top(
             output [`DATA_W-1:0]     tb_uart_rdata,
             output                   tb_uart_ready,
 
-
+            input                    tb_uart_rts
    );
 
    // Wires that will connect both units
-   wire serial_data_to_fpga; // wire from uart going to fpga uart
+   wire serial_data_to_fpga;
```
jjts, line 32: there's no fpga here --- I will merge anyway as this is verilator specific only and can be fixed

PR #118, opened 2021-04-20
update test log files

JDLopes:


jjts:
tests are apssing -- no need to replace logs

PR #139, opened 2021-07-01
improve documentation Makefiles

JDLopes:


jjts:
This seems to be untested changes 

making documents involves running the whole flow

JDLopes:
> This seems to be untested changes
> 
> making documents involves running the whole flow

I left the pd generation untouched, I only improved the presentation generation. So, there is no need for running the whole flow. It may seems that it is untested, but it was tested.

PR #144, opened 2021-07-04
implement lock files when using FPGAs; update test.expected file for xcelium simulator

JDLopes:


jjts:
I dont get the NORUN logic

If it is a NORUN you do nothing 

No relation to lock file

JDLopes:
> I dont get the NORUN logic
> 
> If it is a NORUN you do nothing
> 
> No relation to lock file

Exactly! If NORUN is 1, the if statements filter the lock file behaviour. Just like what happens with the rest of the commands (load, run console, etc).

PR #151, opened 2021-07-19
changed soffice to libreoffice

AndreMerendeira:


JDLopes (commented):
Did you test the commands on terminal? When I tried to run them on my PC, both soffice and libreoffice commands worked, hence my question.

AndreMerendeira:
I tested it on the VM and soffice didn't work. But you're right, soffice and libreoffice both work on my PC and it's the same version as the ubuntu in the VM. Maybe it works for people who had soffice already installed before the name change

PR #170, opened 2021-08-26
solve sim_system_top.v problem: clean was deleting this file; update .gitignore list; cleanup verilator Makefile

JDLopes:
note: verilator testbench has to be finished


jjts:
Verilator will be dealt with separately

PR #171, opened 2021-08-26
add missing dependency to system.v target; uncomment line, so the synthesis timing report show slack value both with and without DDR

JDLopes:


jjts:
fix may have been accidently deleted

use an if statement in the xdc file: if ddr not used include constraint

PR #172, opened 2021-09-03
add core.mk file; use new CORE_NAME variable to create a prefix for peripherals macros, and so, allow to instantiate a SoC as a peripheral of another SoC; make some changes to the software level to support peripherals macros' prefix

JDLopes:


jjts:
i don't like this solution

software/software.mk
```diff
@@ -1,7 +1,10 @@
 defmacro:=-D
 incdir:=-I
+include $(ROOT_DIR)/core.mk
 include $(ROOT_DIR)/system.mk
 
+PREFIX:=$(CORE_NAME)_
```
jjts, line 6: useless renaming, use CORE_NAME directly

software/software.mk
```diff
@@ -25,10 +28,14 @@ HDR=$(SW_DIR)/system.h
 #SRC=$(SW_DIR)/*.c
 
 #peripherals' base addresses
-periphs.h: periphs_tmp.h
+$(SW_DIR)/periphs.h: $(SW_DIR)/periphs_tmp.h
```
jjts, line 31: why change its place ?

software/software.mk
JDLopes: It is not useless. If I didn't use it like this, it just doesn't work because of the inclusion of embedded.mk files in firmware Makefile will overwrite the variable CORE_NAME. So I found this simple workaround.

software/software.mk
JDLopes: I can revert it. It is not a problem. I just thought software.mk file was a better place to put these targets.

.gitignore
```diff
@@ -9,7 +9,7 @@ software/*/*
 !software/*/*.expected
 !software/*/*.py
 
-!software/bootloader/boot.c
+!software/bootloader/boot_core.c
```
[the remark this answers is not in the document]
JDLopes: I didn't remember a better name for the file without the CORE_NAME prefix on peripherals' macros.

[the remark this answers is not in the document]
jjts: Yes, CORE_NAME is teh problem

jjts:
doesn't stick at all

PR #175, opened 2021-09-18
uncomment create_clock in xdc file; add missing dependencies to system.v and system_tb.v targets

JDLopes:


hardware/hardware.mk
```diff
@@ -62,7 +62,7 @@ VSRC+=system.v
 IMAGES=boot.hex firmware.hex
 
 # make system.v with peripherals
-system.v:
+system.v: $(SRC_DIR)/system_core.v
```
jjts, line 65: these dependencies are not useful since system_core.v never changes...

but they do no harm either

JDLopes: Yes, they are. Not every day on every projects, but I just need them for iob-soc-tester project.

PR #176, opened 2021-09-28
update UART submodule

JDLopes:


jjts:
How many times must I say that there NO PRs to master ?

PR #184, opened 2021-10-04
add framework to generate ASIC memories; update .gitignore file

JDLopes:


jjts:
dir/file structure is unnecessary 

we only need a build directory : asic/umc130 for this node 

Inside this build dir we generate the memories, synthesize and later we may do place & route too

verilog sources should be scanned for the memories to implement


Initially this build dir has the set of scripts to do that -- no need to separate them in different directories 

You may, however place the different build artefacts in different subfolders if necessary

PR #196, opened 2021-10-10
Verilator

PedroAntunes178:
Update verilator branch with new Python console.
Two FIFO files are created under the console directory. When passing the argument -L to make the console uses the FIFO files to simulate a serial connection.


Makefile
```diff
@@ -15,6 +15,16 @@ sim-test:
 sim-clean:
 	make -C $(SIM_DIR) clean clean-testlog
 
+#
+# CONSOLE
```
jjts, line 19: console is not run  from top makefile 

it is always run implicitly 

please remove

jjts: make fpga-run triggers console automatically

software/console/console.py
```diff
@@ -0,0 +1,204 @@
+#
```
jjts, line 1: the first line should be 

#!/bin/python3

the script should be named console and not console.py to keep everything else compatible


this allows the program to run by just typing 

./console 

if you are in the console directory or

path/to/console 

from everywhere else

software/console/Makefile
```diff
@@ -2,6 +2,7 @@ ROOT_DIR:=../..
 include $(ROOT_DIR)/system.mk
 
 SERIAL:=/dev/ttyUSB0
+LOCAL:=0
 
 CC:=gcc
```
jjts, line 7: delete ?

software/console/Makefile
```diff
@@ -12,24 +13,24 @@ HDR:=console.c $(UART_DIR)/software $(FIRM_DIR)/UARTsw_reg.h
 
 SRC:=console.c rs232comm.c
```
jjts, line 15: delete the above?

software/console/Makefile
PedroAntunes178: LOCAL indicated that the console was to run with the test bench.
I have now changed the name of the variable to TEST_BENCH.
So when calling "make console TEST_BENCH=1" it knows that there is no serial port and that the program will read and write to fifo files.

software/console/Makefile
PedroAntunes178: Yes. Deleted

software/console/console.py
PedroAntunes178: And used:
#!/bin/env python3

Makefile
PedroAntunes178: Okay, removed.

PR #198, opened 2021-10-11
update .expected files; use --delete flag on clean-remote targets only; fix .gitignore file; minor improvements in FPGAs' Makefiles; merge with iob-soc/master

JDLopes:


Makefile
```diff
@@ -123,7 +123,7 @@ test-doc-clean:
 	make doc-clean DOC=pb
 	make doc-clean DOC=presentation
 
-test: test-clean test-pc-emul test-sim test-fpga test-asic test-doc
+test: test-clean test-pc-emul test-sim test-fpga test-doc
```
jjts, line 126: any test target for asic ?

JDLopes: There is one, but it is not fully functional.

PR #214, opened 2021-10-21
remove .load file in case of a SIGINT to be caught on trap; merge with iob-soc/master

JDLopes:


jjts:
I don't see any need 

If already programmed it should work

PR #228, opened 2021-11-03
Makefile for OpenRAM/OpenLane. OpenRAM make rule added

microSharjeel:
This rule in the Makefile will run the OpenRAM (provided OpenRAM and all its dependencies are installed on that machine). 
Next step is to integrate it with asic.mk 
We can discuss this 
regards
Sharjeel

### DIFF
--- a/hardware/asic/README.md
+++ b/hardware/asic/README.md
@@ -242,6 +242,7 @@ git clone https://github.com/VLSIDA/OpenRAM
   ```bash
   export OPENRAM_HOME="$HOME/openram/compiler"
   export OPENRAM_TECH="$HOME/openram/technology"
+  export OPENRAM_CONFIG="$HOME/openram"
   ```
 3. Also add OPENRAM_HOME to your PYTHONPATH
 ```bash
--- a/hardware/asic/skywater/skywater130.mk
+++ b/hardware/asic/skywater/skywater130.mk
@@ -1,1 +1,7 @@
 #Makefile for OpenLane and OpenRAM started on 2/11/21
+SRAM:
+	python3 $(OPENRAM_HOME)/openram.py $(OPENRAM_CONFIG)/myconfig.py
+	
+.PHONY clean:
+	rm -rf __pycache__
+	rm -rf temp


hardware/asic/skywater/skywater130.mk
```diff
@@ -1 +1,7 @@
 #Makefile for OpenLane and OpenRAM started on 2/11/21
```
jjts, line 1: rename this file and call it simply Makefile

microSharjeel: Ok Sir

hardware/asic/skywater/skywater130.mk
```diff
@@ -1 +1,7 @@
...
+SRAM:
+	python3 $(OPENRAM_HOME)/openram.py $(OPENRAM_CONFIG)/myconfig.py
```
jjts: OPENRAM_HOME should be OPENRAM_COMPILER
OPENRAM_CONFIG should be OPENRAM_HOME

hardware/asic/README.md
```diff
@@ -242,6 +242,7 @@ git clone https://github.com/VLSIDA/OpenRAM
   ```bash
   export OPENRAM_HOME="$HOME/openram/compiler"
```
jjts, line 243: mind the suggested name changes 

you can't assume openram is installed in $HOME


Instead ask the user to 

export OPENRAM_HOME=path/to/openram

so this can be any directory the user may like

hardware/asic/skywater/skywater130.mk
microSharjeel: OPENRAM_HOME is py file for OpenRAM
and OPENRAM_CONFIG has the configuration that user can modify to change all the params of the SRAM
I will modify them now

hardware/asic/skywater/Makefile
```diff
@@ -0,0 +1,7 @@
+#Makefile for OpenLane and OpenRAM started on 2/11/21
+SRAM:
+	python3 $(OPENRAM_HOME)/openram.py $(OPENRAM_CONFIG)/myconfig.py
+	
+.PHONY clean:
```
jjts, line 5: I am not aware of this syntax. Are you sure?

I use 
.PHONY: clean 

let's start putting the phony list before the targets 

there is no target clean, so  pls create it after you declare it as phony

hardware/asic/skywater/Makefile
```diff
@@ -0,0 +1,7 @@
+#Makefile for OpenLane and OpenRAM started on 2/11/21
+SRAM:
```
jjts, line 2: use lower case for target names

hardware/asic/README.md
microSharjeel: Ok I got it sir I am going to change it now

hardware/asic/skywater/Makefile
microSharjeel: Sir I am making all these changes now

microSharjeel: Kindly check now

PR #235, opened 2021-11-07
Verilator

PedroAntunes178:
Added cocotb simulation.
Still not working tho,

Had to changed system_core_tb.v in order to adapt to the python testbench.
The error might be on system_core_tb.v


hardware/simulation/cocotb/cocotb.mk
```diff
@@ -0,0 +1,20 @@
+#Makefile
```
jjts, line 1: Why is this not part of the Makefile? They are in the same directory. Whi file will include this makefile segment ?

PedroAntunes178: The idea was that cocotb.mk would be called from the Makefile.
The Makefile is called with "make sim" like one would normaly do.
This way there are no target mixing from the original makefiles and the cocotb makefile included.

jjts: I don't understand well but will merge and then we discuss. The Makefile is isolated in the coco_tb directory . Pasting coco_tb.mk would be equivalent I think.

PR #239, opened 2021-11-12
Changes in Makefile for OpenLane Docker start

microSharjeel:


microSharjeel:
Export OPENLANE_MK = <path to OpenLane root folder>

microSharjeel:
I have commented include asic.mk in Makefile. This is because if I include it it says that at line no 9 in asic.mk 
 hardware/hardware.mk  ...... no such file or directory...........

PR #241, opened 2021-11-13
add missing MEM_DIR path

JDLopes:


jjts:
The idea is to get MEM_DIR when cache hardware.mk is added, and not add it explicitly. Why is this not happening?

JDLopes:
> The idea is to get MEM_DIR when cache hardware.mk is added, and not add it explicitly. Why is this not happening?

Oh, sorry! I will fix it.

PR #263, opened 2021-12-01
Makefile src files issue resolved 

microSharjeel:
Hi Sir,
I have resolved the Makefile issue of src files. Now we will add the config.tcl file to the Skywater folder and Makfefile will copy it to the designs/system_core folder. User can change config.tcl to add any other option as specified in README.md. 
Now I am working on memory generation. Build.sh is there in the folder, but it may or may not be required, I will decide it after the sram macro placement is added to the Makefile
One more thing is that the name of the module in system_core.v file has been changed to system_core. Since OpenLane requires the name of file and module to be the same. 
regards

### DIFF
--- a/hardware/asic/skywater/Makefile
+++ b/hardware/asic/skywater/Makefile
@@ -22,13 +22,13 @@ sram: asic_config.py
 
 # ASIC generation
 asic:  $(VSRC) $(VHDR)
-	mkdir -p $(OPENLANE_DESIGNS)/system_core/src 
-	mkdir -p $(OPENLANE_DESIGNS)/system_core/inc 
-	cp $(VSRC) $(OPENLANE_DESIGNS)/system_core/src 
-	cp $(VHDR) $(OPENLANE_DESIGNS)/system_core/inc 
-	cp config.tcl $(OPENLANE_DESIGNS)/system_core
+	mkdir -p $(OPENLANE_DESIGNS)/system/src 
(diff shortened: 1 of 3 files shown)


jjts:
You can't use system_core.v. This file is only used to help generate system.v, which is the one used, not meant to be synthesized.

Please see in the makefile how syetm.v is generated from system_core.v, which is not by itself a valid source.

microSharjeel:
I have reverted the changes

microSharjeel:
Ok I will change the design name in Makefile.
Actually I thought that it is the top_level module
OpenLane does not use this file
it uses system.v 
so sorry for this change i have reverted the changes
regards

microSharjeel:
Now there are no changes in any iob-soc src file 
so you can merge this if anything else is ok
regards

microSharjeel:
I only use $(VSRC) and $(VHDR) as src and header in Makefile  and not any file separately 
This was just a mistake I have reverted it

microSharjeel:
Ok it is confirmed that OpenLane does not use system_core.v and it uses system.v 
Sorry My mistake

jjts:
This is totally wrong, sorry. You can't just edit the file. You need to first understand how the system is built and then submit a meaningful PR.

jjts:
I will close this PR as it is. PLease submit a new one or ask which files to use as you clearly don't know.

PR #272, opened 2021-12-07
Fixed a typo in figures and presentation. make test-doc now uses resources with DDR=0

AndreMerendeira:


Makefile
```diff
@@ -121,6 +121,10 @@ test-asic-clean:
 	make asic-clean ASIC_NODE=umc130
 
 test-doc:
+	make fpga-clean BOARD=CYCLONEV-GT-DK
```
jjts, line 124: why not clean-all and buld-all?

PR #276, opened 2021-12-10
add variable to use single-port RAM instead of dual-port RAM; update Makefiles; update CACHE and MEM submodules

JDLopes:


jjts:
this type of cahnge DOES not go to master , a branch will be created

PR #277, opened 2021-12-10
add variable to use single-port RAM instead of dual-port RAM; update Makefiles; update CACHE and MEM submodules

JDLopes:


jjts:
I don't understand why you have USE_SPRAM everywhere 

It should be in hardware.mk or config.mk

If we want iob-soc to use SP_RAM than both simulation and FPGA should have that so we can test


This PR cannot be accepted without a test case activating this variable

JDLopes:
> I don't understand why you have USE_SPRAM everywhere
> 
It is everywhere because different technologies can USE_SPRAM equal to 1 or to 0. For now, I decided to put it defined to 0 as default everywhere, except in the SkyWater Makefile. It is defined using ?=, so it can be overwritten for other tests

> It should be in hardware.mk or config.mk
> 
It is added to DEFINE list in hardware.mk, so any SoC platform (simulation, FPGA or ASIC) can use it.

> If we want iob-soc to use SP_RAM than both simulation and FPGA should have that so we can test
> 
> This PR cannot be accepted without a test case activating this variable
I can add a test case for both simulation and FPGA.

PR #287, opened 2021-12-23
Asic

microSharjeel:
Hi
Makefile changes that synthesize the design and next step is STA pass. Also changes in UART for adding UARTsw_reg.v in VHDR 
kindly  have a look

### DIFF
--- a/hardware/asic/skywater/openram_config.py
+++ b/hardware/asic/skywater/openram_config.py
@@ -4,9 +4,10 @@ word_size = 32
 num_rw_ports = 1
 num_r_ports = 0
 num_w_ports = 0
-
+#write size for adding byte enable for sram
+write_size =8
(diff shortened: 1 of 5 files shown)


hardware/asic/skywater/openram_config.py
```diff
@@ -4,9 +4,10 @@
 num_rw_ports = 1
 num_r_ports = 0
 num_w_ports = 0
-
+#write size for adding byte enable for sram
+write_size =8
 # Number of words in the memory
-num_words = 64
+num_words = 32
```
jjts, line 10: why 32 words ? This is possibly not enough for hello world

microSharjeel: this is for test only I will change it as the flow progresses

PR #289, opened 2022-01-04
Changes to synthesize iob-soc with RAM as black box 

microSharjeel:
Hi
This commit has changes to synthesize the design with RAM as black box. The design synthesize properly but a dummy module for black box is required. This is a OpenLane requirement. I will add that in memakerwrap.py and will commit it separately in iob-mem.
I have temporarily added boot.hex file to skywater folder for synthesis purpose 
This is intermediate commit and I am still working on synthesis / STA 
Kindly have a look 
regards


config.mk
```diff
@@ -12,10 +12,10 @@
 FIRM_ADDR_W ?=15
 
 #SRAM SIZE (LOG2)
-SRAM_ADDR_W ?=15
```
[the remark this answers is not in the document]
microSharjeel: I will restore it

jjts:
I do not understand this PR:

1) a long time has passed since the last one so things are forgotten

2) wasn't RAM already done? I thought the problem was ROM 

3) Is it possible or not to do dual-port RAM with openRAM ?

4) Is it possible or not to do ROM with opnRAM?

microSharjeel:
No RAM wrapper was done but it requires synthesis and STA and Macro Placement steps to complete successfully This commit is at the stage of STA completed successfully but we only provide the wrapper for RAM and the instance of RAM inside the wrapper is not in VSRC and does not even exist .... so a black box .... and will be replaced by the OpenRAM block
For dual port RAM I asked at OpenRAM forum but right now there is issue in it and only SP is possible

hardware/asic/skywater/Makefile
```diff
@@ -80,10 +83,10 @@ clean_sram:
 
 clean_asic:
 	rm -rf $(OPENLANE_DESIGNS)/system
-	rm -rf iob_dp_ram.v
+	rm -rf iob_sp_ram_be.v
 	rm -rf $(OPENLANE_HOME)/run_interactive.tcl
-clean_iob_dp_ram:
```
JDLopes, line 85: I don't understand why have different targets for clean memory wrapper files. Why not clean them in clean_asic target?

microSharjeel:
Yes it is possible to do ROM aswell
and this is the next step

hardware/asic/skywater/Makefile
```diff
@@ -31,16 +32,17 @@ sram: openram_config.py
 	mv temp sram
 #SRAM wrapper generation
 
-iob_dp_ram.v:
+iob_sp_ram_be.v:
 #It was added for test only -will be removed-
 #	~/Documents/iob-mem/software/python/memakerwrap.py $(ASIC_MEM_TECH) iob_dp_ram $(ASIC_DPRAM_TYPE) 0 1 $(SRAM_W) 8 4 1 > $@
-	$(MEM_DIR)/software/python/memakerwrap.py $(ASIC_MEM_TECH) iob_dp_ram $(ASIC_DPRAM_TYPE) 0 1 $(SRAM_W) 8 4 1 > $@
+	$(MEM_DIR)/software/python/memakerwrap.py $(ASIC_MEM_TECH) $(ASIC_MEM_MODULE) $(ASIC_SPRAM_TYPE) 0 1 $(SRAM_W) 8 4 1 > $@
```
JDLopes, line 38: There is no point in have ASIC_MEM_MODULE variable. You can have it hardcoded as before, it is only used in here.

hardware/asic/skywater/Makefile
microSharjeel: I will make one target

hardware/asic/skywater/Makefile
microSharjeel: but this variable is only here in this Makefile

hardware/asic/skywater/Makefile
microSharjeel: I will revert the config.mk file changes

hardware/asic/skywater/Makefile
microSharjeel: And boot.hex is there only temporarily for testing only i mentioned it in original message

JDLopes: > but this variable is only here in this Makefile

That's why you don't need to have a variable. You can write it hardcoded here.

microSharjeel: I will do it

JDLopes: > And boot.hex is there only temporarily for testing only i mentioned it in original message

I understand that, but boot.hex is a generated file. So you should let the Makefiles generate it automatically and use it as is.

PR #290, opened 2022-01-11
Asic

microSharjeel:
Hi,
Changes in skywater files for manual macro placement of RAM macro.
regards,


config.mk
```diff
@@ -12,10 +12,10 @@
 FIRM_ADDR_W ?=15
 
 #SRAM SIZE (LOG2)
-SRAM_ADDR_W ?=15
+SRAM_ADDR_W ?=7
```
JDLopes, line 15: Why do you need to change this variable? The firmware uses more than 2**7 words of memory. If you want to try new values for debugging you should call the make command with those variables set, like so, 'make asic-synth SRAM_ADDR_W=7'. In any case, SRAM_ADDR_W set with 7 is not useful. Please revert this change.

config.mk
```diff
@@ -12,10 +12,10 @@
...
 
 #RAM TYPE (single-port or dual-port)
-USE_SPRAM ?=0
+USE_SPRAM ?=1
```
JDLopes, line 18: Same here.

hardware/asic/skywater/Makefile
```diff
@@ -50,23 +51,27 @@ ADD:   DEL
...
 	$(MKDIR) $(OPENLANE_DESIGNS)/system/macros/lef
 	$(MKDIR) $(OPENLANE_DESIGNS)/system/macros/gds
 	cp  $(wildcard sram/*.lef) $(OPENLANE_DESIGNS)/system/macros/lef 	
 	cp  $(wildcard sram/*.gds) $(OPENLANE_DESIGNS)/system/macros/gds 		
 	cp  $(wildcard temp/*.v) $(OPENLANE_DESIGNS)/system/src
 	cp  $(wildcard temp/*.vh) $(OPENLANE_DESIGNS)/system/src
+	cp  $(wildcard temp/boot.hex) $(OPENLANE_DESIGNS)/system/src
 	cp run_interactive.tcl $(OPENLANE_HOME)
 	cp config.tcl $(OPENLANE_DESIGNS)/system
 	cd $(OPENLANE_HOME);
```
JDLopes, line 68: Don't lease empty lines in a Makefile recipe. And do not commit commented code lines.

hardware/asic/skywater/config.tcl
```diff
@@ -7,7 +7,8 @@ set ::env(VERILOG_FILES) "\
     designs/system/src/system.v\
     designs/system/src/boot_ctr.v\
     designs/system/src/int_mem.v\
-    designs/system/src/iob_dp_ram.v\
+    designs/system/src/iob_sp_ram_be.v\
```
JDLopes, line 10: You should try to pass a list of VSRC files instead of having a hardcoded list.

hardware/asic/skywater/Makefile
```diff
@@ -1,46 +1,47 @@
 ROOT_DIR=../../..
-
+include ../asic.mk
 OPENLANE_DESIGNS:=$(OPENLANE_HOME)/designs
 IMAGE_NAME ?= efabless/openlane:2021.07.29_04.49.46 #latest
 ASIC_NODE=skywater
-#defmacro := -D
+USE_SPRAM = 1
```
JDLopes, line 6: You don't need these 3 variables here since you include asic.mk file, that includes hardware.mk, that includes config.mk file. They are already set at this point.

JDLopes (changes requested):

config.mk
microSharjeel: I reverted it to its original value
I changed it earlier but as it was pointed out in last PR I reverted all things in config.mk
this is the revert

config.mk
microSharjeel: same reply this is the revert

hardware/asic/skywater/Makefile
microSharjeel: is there anything commented here ?

microSharjeel: There is an addition of one necessary line

hardware/asic/skywater/config.tcl
microSharjeel: how I will pass the VSRC to config.tcl file ?

hardware/asic/skywater/Makefile
```diff
@@ -50,23 +51,27 @@ ADD:   DEL
...
 	cp  $(wildcard temp/*.v) $(OPENLANE_DESIGNS)/system/src
 	cp  $(wildcard temp/*.vh) $(OPENLANE_DESIGNS)/system/src
+	cp  $(wildcard temp/boot.hex) $(OPENLANE_DESIGNS)/system/src
 	cp run_interactive.tcl $(OPENLANE_HOME)
 	cp config.tcl $(OPENLANE_DESIGNS)/system
 	cd $(OPENLANE_HOME);
 
        # For non-interactive flow of OpenLane by using iob-soc SRAM and macro placement config file -please do not remove-
 #$(DOK_CMD) sh -c   "./flow.tcl -design system  -p "read_verilog -I/$(OPENLANE_DESIGNS)/system/inc" -tag soc -overwrite -config file /system/config.tcl"
-#For interactive flow for SRAM Macro placement
```
JDLopes, line 68: is there anything commented here ? > Yes it is!

hardware/asic/skywater/Makefile
microSharjeel: Sorry but USE_SPRAM was reverted to its original value after you and jose said me to set any value to this in my Makefile this is what I did as you both said

config.mk
JDLopes: iob-soc/ master project has a default value for the SRAM_ADDR_W of 15.

hardware/asic/skywater/Makefile
microSharjeel: I need USE_SPRAM =1 and in config,mk it is 0 by default then where I can set them? other than Makefile

hardware/asic/skywater/Makefile
JDLopes: It should be right next to these ones. Why the empty line with no target and then an indented line?

hardware/asic/skywater/config.tcl
JDLopes: Try to see how it is done for the .tcl script for FPGAs and umc130. Those platforms use .tcl scripts that read VSRC as an argument.

jjts:
There is only one Makefile that you are working on. The one in your directory.

config.mk is a makefile segment that is included automatically in every makefile 

If you want to include something in your Makefile use it or any other makefile segment that you own and include. Don't use a makefile segment that is used by many other makefiles as this will disrupt the other makefiles.

Need to better understand the Makefile structure.

hardware/asic/skywater/Makefile
JDLopes: You can write the following command in the terminal 'make asic-synth USE_SPRAM=1'.

jjts:
Nothing is reverted -- a PR cannot show changes to config.mk.

I will close this PR. Fix your files and resubmit a clean one.

If you don't know how to revert stuff with git you need to learn:

git reset <hash>

<hash> is a commit number that does not yet have changes to config.mk 

git status 

see the changes relative to the commit you reset to

edit files at your will 

you can discard the edits to config.mk with 

git checkout config.mk 

will restore the version present at <hash>

resubmit the PR as this is the second time I warn you that changes to config.mk are not acceptable.

Unjustified comments in the code are also not acceptable 

We need to all drive on the same side of the road or otherwise this becomes a mess

jjts:
I mean git reset "hash"

jjts:
will restore the version present at "hash"

PR #292, opened 2022-01-19
Corrected PR

microSharjeel:
Hi,
I have tried to incorporate all previous observations in it kindly have a look
best regards


hardware/asic/skywater/Makefile
```diff
@@ -1,46 +1,35 @@
...
-
-VSRC+=iob_dp_ram.v
-
-# Memory technology
-ASIC_MEM_TECH:=skywater130
-
-# Memory type
-ASIC_DPRAM_TYPE:=spram
-
+VSRC+=iob_sp_ram_be.v
```
jjts, line 8: are these the wrappers? what have you done with the others that were in VSRC?

hardware/asic/skywater/config.tcl
```diff
@@ -1,36 +1,32 @@
 # User config
-
 set ::env(DESIGN_NAME) system
-
 # Change if needed
 set ::env(VERILOG_FILES) "\
```
jjts, line 4: do a get ::env of VSr first and yo don't need to list manually

hardware/asic/skywater/Makefile
microSharjeel: this is the wrapper that is generated by the Makefile by calling memakerwrap.py

hardware/asic/skywater/config.tcl
microSharjeel: Will VSRC give me all the paths for source file from inside the Openlane/designs/src. 
This statement just tells OpenLane that these are the verilog source files. These all paths for all source files have to be " designs/my _design/src " . VSRC have all the paths from iob-soc which OpenLane does not recognize. 
Still I can do  designs/src/*.v to tell the OpenLane about all the source files and it will not be hardwired but as I know there are files like gen_if.v which is actually a header file and if i do a *.v then OpenLane treats gen_if.v as verilog source file and throws error. So these are the reasons for manual list

microSharjeel: what we can do is to make a list of all the files that are there in src folder in OpenLane/designs/my_design and pass it to config.tcl
This list will be updated when vsrc will change and there will be no manual list
this is my suggestion
regards

hardware/asic/skywater/Makefile
microSharjeel: I think with asic=1 in asic.mk we do not get wrappers in VSRC and they are generated and added by skywater Makefile

PR #313, opened 2022-02-14
Verilator

PedroAntunes178:
NOT WORKING AT 100%
- Python console program modified to work with files.
- Cocotb testbench was adapted to work with files instead of FIFOs. Currently, with INIT_MEM=1 it works, but with INIT_MEM=0 it does not work, since it gets blocked after a file is written from console to SoC. Do not know if it is caused by the testbench or the console.
- Started Icarus testbench. Currently, the testbench runs, but it gets in an infinite loop when reading the rxread_reg and txread_reg. Since, they are never returned with the value 1 from cpu_uartread.
- Started Verilator testbench. But, it is not currently the main focus.


jjts:
We are not continuing with Cocotb -- it's too slow

Only icarus and verilator

PedroAntunes178:
> We are not continuing with Cocotb -- it's too slow
> 
> Only icarus and verilator

Used cocotb because it was the fastest way to know if the console program worked with files and easier to detect bugs. Since it is written in python. Additionally, there was already a base testbench working with FIFOs which made it easier to make the change. Contrarily to the other simulators. Furthermore, since the cocotb already worked partly, the Verilog testbench follows the same logic.

PR #316, opened 2022-02-16
merge with iob-soc/devel

microSharjeel:


jjts:
You are targetting the wrong branch

We need branch asic to be up to date with devel and not the other way round

PR #327, opened 2022-02-26
changes to pass test.log and expected in Jenkins

microSharjeel:


jjts:
These changes are rejected -- this is not how we check if tests pass. We can do this later.

The greatest problem of this branch is that it does not do anything related to asic.

make asic-synth should be made to work first

PR #334, opened 2022-03-07
Verilator

PedroAntunes178:
Verification runs with cache. Warning is still displayed, but the simulation no longer exits due to warnings.

There is a problem, when just running "make sim SIMULATOR=verilator RUN_EXTMEM=1" the simulation ends with a CPU trap condition. This does not occur in the old simulation. 
Running "make sim SIMULATOR=verilator RUN_EXTMEM=1 INIT_MEM=0" works as expected.

Probably, the console, when using the cache, always has to be called with the "-f" flag in order to send the firmware binary file. Just like it happens with INIT_MEM=0.  I have to check if this is true and then make the changes for it to happen.

### DIFF
--- a/hardware/simulation/icarus/Makefile
+++ b/hardware/simulation/icarus/Makefile
@@ -4,15 +4,16 @@ defmacro:=-D
 incdir:=-I
 
 CONSOLE_CMD=../../../software/console/console -L
-ifeq ($(INIT_MEM),0)
-	CONSOLE_CMD+=-f
-endif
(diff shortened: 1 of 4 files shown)


PedroAntunes178:
RUN_EXTMEM=1 indeed sets INIT_MEM=0
solved the problem by including the simulation.mk before adding the "-f" flag to the command that calls the console. But I think a better solution might be to call the Makefile under the console directory.

PR #348, opened 2022-03-12
changes to remove OpeRAM dependency

microSharjeel:


jjts:
The gds file is a 0.5kByte memory, right ?

why do we need a gds2 file (layout) for synthesis? for the time being we just want synthesis to complete without errors.

After that post-synth simulation.

Only at a later stage we will need the layout.

It is premature to merge gds2 files.

jjts:
By that time maybe OpenRAM can be used and no need to merge a 3MB file!

microSharjeel:
This gds and lef files are OpenRAM generated RAM files. Combined lef + gds forms the hardened macro of the SRAM. They are used in the floorplan step to place this macro in the black box. 
If we only require synthesis and STA and this time then I will close this PR and submit another in which flow is only up to synthesis

PR #350, opened 2022-03-12
corrected cp command

microSharjeel:


hardware/asic/skywater/Makefile
```diff
@@ -40,7 +40,7 @@ $(DEFINE): sw boot.hex
 	cp boot.hex temp_src
 
 synth:  clean $(VSRC) $(VHDR) temp_src temp_inc $(DEFINE) 
-	cp /opt/OpenLane . 
+	cp -r /opt/OpenLane .
```
jjts, line 43: cannot have have hardcoded paths as users may use different paths

pls call it OPENLANE_SRC_DIR

jjts:
do you have more hardcoded stuff ?

jjts:
also, change the readme so users set the right env variables

hardware/asic/skywater/Makefile
microSharjeel: OPENLANE_SRC_DIR 
should it be inside Makefile only or it should be in bash?

microSharjeel:
no more hardcoded stuff i believe

hardware/asic/skywater/Makefile
jjts: how can it be in makefile without being hardcoded?

jjts: the user must define it -- write that in the readme

microSharjeel:
In README it is already generalized like see following statements
export OPENLANE_HOME=path/to/OpenLane/root
export PDK_ROOT=$OPENLANE_HOME/pdks

hardware/asic/skywater/Makefile
microSharjeel: export OPENLANE_SRC_DIR=<path to openlane src directory>
I have added to README
and this will be in bash now

microSharjeel:
check now changes in README and Makefile now appear in PR

PR #353, opened 2022-03-12
change in OPENLANE_HOME path

microSharjeel:
This change in path has solved this problem this complete path is requirement of the doclker, 
Please remove OPENLANE_HOME from bashrc since it is now in Makefile


jjts:
cannot accept hardcoded paths 

second time 

this path does not even exist in my setup

only in yours

PR #362, opened 2022-03-19
Changes to get docker usr_id dynamically and OpenLane copy issue solved

microSharjeel:


microSharjeel:
Hi
This PR solves the docker usr_id issue. Now the Makefile will get the port ID dynamically by which docker connects to host system. Also the OpenLane will only be copied if it is not found locally in the skywater folder. Clean target will only clean the OpenLane/designs/system folder since this is design folder created by OpenLane every time it is run.

PR #367, opened 2022-03-22
fix(console): usage error, update SERIAL

P-Miranda:
- Update SERIAL path, note that this now assumes a SYMLINK to
  `/dev/usb-uart`.
    - This fixes problems related to the same device beeing attributed a
      random value in `/dev/ttyUSBx` that can vary from boot to boot
- Fix `usage()` function
    - previous implementation called `cnsl_perror()` twice:
        - 1st time prints usage
        - 2nd time prints error message
    - however `cnsl_perror()` calls `exit(1)`, so the error message is
      never printed
    - Now `usage()` prints usage message and calls `cnsl_perror()`

### DIFF
--- a/software/console/Makefile
+++ b/software/console/Makefile
@@ -1,6 +1,7 @@
 ROOT_DIR:=../..
 include $(ROOT_DIR)/config.mk
 
+# Need to add udev rule to have this symlink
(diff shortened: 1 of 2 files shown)


P-Miranda:
PR is redundant.
Fixes included in PR #370

PR #375, opened 2022-03-24
remove unused signal in top_system.v file for AES-KU040-DB-G FPGA

JDLopes:


jjts:
I will not accept these mini PRs anymore 

Please get it to work and then do a meaningful PR

PR #389, opened 2022-04-27
add firmware.bin dependency when run console; fix documents test targets; add general clean target

JDLopes:


hardware/simulation/simulation.mk
```diff
@@ -63,7 +63,7 @@ else
 	bash -c "trap 'make kill-remote-sim' INT TERM KILL; ssh $(SIM_SSH_FLAGS) $(SIM_USER)@$(SIM_SERVER) 'make -C $(REMOTE_ROOT_DIR) sim-build SIMULATOR=$(SIMULATOR) INIT_MEM=$(INIT_MEM) USE_DDR=$(USE_DDR) RUN_EXTMEM=$(RUN_EXTMEM) VCD=$(VCD) TEST_LOG=\"$(TEST_LOG)\"'"
 endif
 
-run:
+run: firmware.bin
```
jjts, line 66: No. When you run simulation the fw is already built. No cross dependencies.

jjts:
pls try again. Closing

hardware/simulation/simulation.mk
JDLopes: Yes, it is already built. But this dependency is here only for copying the firmware.bin to the build folder (icarus, CYCLONEV, etc).

PR #393, opened 2022-05-02
Minor improvements

JDLopes:
- Manage to support all simulators with the same waves file;
 - Open a new gtkwave process if there is none opened yet;
 - Do not transfer dummy log files when using FPGAs;
 - Remove unused include from firmware.c file.


hardware/simulation/verilog_tb/system_core_tb.v
```diff
@@ -5,7 +5,7 @@
 
 //PHEADER
 
-module system_tb;
+module TOP;
```
jjts, line 8: I dont get this

JDLopes: This change allows the user to use the same waves file on Icarus and Verilator. I tried to change the path on the Verilator side, but it seems to be somehow hardcoded. It always starts on the TOP. So I changed it on the Icarus side. Now the user can open the VCD generated with any simulator using only a single waves file

hardware/fpga/fpga.mk
```diff
@@ -39,7 +39,9 @@ else
 	ssh $(BOARD_USER)@$(BOARD_SERVER) "if [ ! -d $(REMOTE_ROOT_DIR) ]; then mkdir -p $(REMOTE_ROOT_DIR); fi"
 	rsync -avz --delete --force --exclude .git $(ROOT_DIR) $(BOARD_USER)@$(BOARD_SERVER):$(REMOTE_ROOT_DIR) 
 	bash -c "trap 'make queue-out-remote' INT TERM KILL; ssh $(BOARD_USER)@$(BOARD_SERVER) 'make -C $(REMOTE_ROOT_DIR)/hardware/fpga/$(TOOL)/$(BOARD) $@ INIT_MEM=$(INIT_MEM) FORCE=$(FORCE) TEST_LOG=\"$(TEST_LOG)\"'"
-	scp -r $(BOARD_USER)@$(BOARD_SERVER):$(REMOTE_ROOT_DIR)/hardware/fpga/$(TOOL)/$(BOARD)/*.log .
+ifneq ($(TEST_LOG),)
+	scp $(BOARD_USER)@$(BOARD_SERVER):$(REMOTE_ROOT_DIR)/hardware/fpga/$(TOOL)/$(BOARD)/test.log .
+endif
```
P-Miranda, line 44: I changes this to play nice with the ethernet submodule.
When the ethernet is present we want to get a `soc.log` and a `ethernet.log` from the remote machine.
My addition is a by hacky, I'll try to find some better solution.

Also, we should use `$(TEST_LOG)` variable to get the name of the .log file that we want:
```Make
# Assume TEST_LOG ="> logname.log"
scp $(BOARD_USER)@$(BOARD_SERVER):$(REMOTE_ROOT_DIR)/hardware/fpga/$(TOOL)/$(BOARD)/$(lastword $(TEST_LOG)) .
```

PR #402, opened 2022-05-25
changes to run post synth simulation in skywater

microSharjeel:


jjts:
I though memories need to be at the top-level but they are still buried in sram.v for example ? Why? Can place and route be done this way ?

microSharjeel:
> I though memories need to be at the top-level but they are still buried in sram.v for example ? Why? Can place and route be done this way ?

Yes macro placement can be done at any level. What is required is to pass the hierarchical name to macro placement of OpenLane.

microSharjeel:
Yes macro placement can be done at any level. What is required is to pass the hierarchical name to macro placement of OpenLane.

jjts:
Is there any change to this PR to address the comments? Please close it.

PR #418, opened 2022-06-17
Enable support for multiple clk's in the verilator test-bench; Update gitignore file;

PedroAntunes178:
In the verilator .cpp test-bench, the Timer function was changed. It now receives the time that is supposed to pass in ns instead of the number of half cycles. This allows for it to support multiple clk signals.


jjts:
what is the real need for this change? Why only in Verilator? Only one test has failed, it should be ok

PR #424, opened 2022-07-02
Update LIB submodule; Add feature to console; Use fixed Periphral and Boot bit; Add split between internal and external data bus in system_core;

PedroAntunes178:


jjts:
I see console modifications are here too...

Ok, I see the problem: console is a tool we use in multiple iob-soc-* repositories.

It is simply not possible to update console in all of them.

I suggest we move console to iob-lib. After all is just one more python script.

jjts:
Use fixed Peripheral and Boot bit; Add split between internal and external data bus in system_core; -- these are very interesting updates.

Please move Console to iob-lib, do a PR there, and review this PR to make it work with the new console location.

PedroAntunes178:
> Please move Console to iob-lib, do a PR there, and review this PR to make it work with the new console location.

Should the console submodule be in the python scripts directory? Or should there be a console folder (containing only the console script) under LIB/software?

jjts:
Console is not a submodule (why do you say so?)

console can be put in iob-lib's software/python directory

PedroAntunes178:
> Console is not a submodule (why do you say so?)
I meant software, sorry :)

> console can be put in iob-lib's software/python directory
Ok, the PYTHON_DIR is defined in hardware.mk, I was thinking of moving the define to the config.mk.

jjts:
Console is not yet in iob-lib?

PR #432, opened 2022-07-26
Simplify reset sequence; Fix not showing reset trace;

PedroAntunes178:


PedroAntunes178:
Calling `Timer(100);` makes so that 100 ns pass with reset value at 1;

PR #439, opened 2022-08-02
 uncommented the tcl file pin locations and assignments and fixed an input name in verilog file.

behzadjoudat:


hardware/fpga/quartus/DE5-NET/verilog/top_system.v
```diff
@@ -2,7 +2,7 @@
 `include "system.vh"
 
 module top_system(
-  input        IE_PERST_n,
+  input        PCIE_PERST_n,
```
jjts, line 5: You should not have removed resetn nor replaced it with PCIE_PERST_n. If this signal is an input, it does not matter if temporarily it isn't connected internally: it will when the PCIe interface core is added

behzadjoudat: fixed.

PR #440, opened 2022-08-02
the device name changed in the tcl file 

behzadjoudat:
also the pcie qsys folder added with it's contents(in development)


jjts:
from the qsys directory I only need the .qsys file 

Don't you check the added files before committing?

PR #444, opened 2022-08-22
 the errors in qsys are fixed within the quartus app and axi_bridge is also added next to the pcie IP

behzadjoudat:


jjts:
trust this is good or at least progress

PR #453, opened 2022-10-01
all the documentation for de5-net and design samples as well as the c…

behzadjoudat:
…hanges we did for the uart


jjts (commented):
Pls close.

PR #455, opened 2022-10-02
the uart works ok with the de5-net with the changes in the baudrate

behzadjoudat:


jjts:
This PR includes a LIB update where console has been changed but no PR for lib has been done.

You need to do a PR in the main iob-lib branch before you can submit this PR.

PR #456, opened 2022-10-03
the baudrate change target added to the makefile and the de5-net uart works with maximum of 19200 baudrate

behzadjoudat:


jjts:
This is PR is totally unacceptable

PR #459, opened 2022-10-04
the pcie peripheral dir added

behzadjoudat:


jjts:
Since when do we add peripherals directly in iob-soc? Peripherals must be reusable in other systems. I created iob-pcie for that purpose. Use it.

PR #466, opened 2022-10-13
the pcie just sends one number and gets it back

behzadjoudat:
As it was with the chnl tester it was working fine but after connecting the wires to the peripheral it works fine with sending and receiving one packet number but not with more than that. trying to figure out why.

the rescan also added to the prog.sh so after programming it rescans the bus for the pcie device be recognized by the riffa

### DIFF
--- a/hardware/fpga/quartus/DE5-NET/verilog/top_system.v
+++ b/hardware/fpga/quartus/DE5-NET/verilog/top_system.v
@@ -293,43 +293,6 @@ module top_system
          .CHNL_TX_DATA_VALID            (chnl_tx_data_valid[C_NUM_CHNL-1:0])
 	 );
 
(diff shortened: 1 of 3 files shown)


hardware/fpga/quartus/DE5-NET/verilog/top_system.v
```diff
@@ -368,41 +331,86 @@ module top_system
        rst_cnt <= rst_cnt - 1'b1;
 
    assign sys_rst  = (rst_cnt != 16'h0);
+/*
 
-   
+       genvar                                   i;
```
jjts, line 336: why is this at the top-level? I thought the channel tester is no longer needed if we connect riffa to pcie peripheral

PR #469, opened 2022-10-18
The final working state with riffa and pcie periph

behzadjoudat:
the test code implemented is the channel tester of the riffa driver itself in the pcie periph dir which is to be chenged with the design of the user using the existing pios

### DIFF
--- a/hardware/fpga/quartus/DE5-NET/verilog/top_system.v
+++ b/hardware/fpga/quartus/DE5-NET/verilog/top_system.v
@@ -293,43 +293,6 @@ module top_system
          .CHNL_TX_DATA_VALID            (chnl_tx_data_valid[C_NUM_CHNL-1:0])
 	 );
 
-       genvar                                   i;
-    generate
-        for (i = 0; i < C_NUM_CHNL; i = i + 1) begin : test_channels
-            // Instantiate and assign modules to RIFFA channels. Users should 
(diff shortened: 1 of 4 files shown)


behzadjoudat:
the reason the check fails is that it tries to simulate all the .v files that are also in the riffa dir which it should not do that as they are there just to be included by some files. 
for example in the icarus failed test it indicates the missing vh file in the wrapper code which is included for the making of the fpga and not for the simulation.

also if you merge the iob-pcie pr then I will update this one with the updated status. you don't get notified about that repo I think.

jjts:
I can't merge this because iob_pcie.v is full of serious mistakes.

It is pointless if something is working if customers will reject it. They will not take clunkiness.

jjts:
I have fixed the private submodules and the checks are still failing 

A PR is only merged if checks pass 

Is your repo up to date with iob-soc / master? Do you know how to do that?

behzadjoudat:
I have explained the reason the check fails in a comment before yours. but let me see what can I do to fix relevant errors.

jjts:
you explained and I fixed it 

but there are more fails 

click  the red crosses of the failed tests

you should be able to figure out what went wrong a suggest a fix 

File .github/workflows/ci.yml specifies the tests that are run. 

You can also run the same tests on your terminal and figure it out 

First of all, can you figure out what each test does?

PR #475, opened 2022-11-12
Update LIB submodule to master HEAD;

PedroAntunes178:
Had to add the `param AXI_LEN_W = 8;` to the simulation top file, fpga's top files, system_core and ext_mem Verilog modules.


jjts:
cyclone test failed, can you see why?

PedroAntunes178:
I think it might be that  `localparam AXI_LEN_W = 8;` in https://github.com/IObundle/iob-soc/blob/84e51a4a15486236f4749b3ef9369d7530e179cf/hardware/fpga/quartus/CYCLONEV-GT-DK/verilog/top_system.v#L38 should be 4 instead of 8.
But I am not sure.
I update the PR and see if it passes the tests.

PR #478, opened 2022-12-01
feat(build-lib): build dir generates successfully

arturum1:
- Build dir generates successfully with `make setup`, however, the system does not compile in simulation.
  - It fails simulation due to missing iob_intercon.vh from LIB. Need to
    find alternative or modify Verilog code to not use it.
- Add sim and doc flows
  - fpga flow currently has an issue
- Select fixed CACHE_CONFIG to 'iob'. This may need to change to 'axi'?
- Rename axi_gen.py to if_gen.py (according to new LIB)
- Add iob_soc_setup.py
- Update submodules

### DIFF
--- a/config_setup.mk
+++ b/config_setup.mk
@@ -6,7 +6,7 @@ VERSION=0070
 DOC_RESULTS=
 
 #supported flows
-FLOWS := pc-emul
+FLOWS := pc-emul sim doc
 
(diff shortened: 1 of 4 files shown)


jjts:
Intercon.vh macros are in iob_lib.vh now

PR #479, opened 2022-12-05
feat(build-lib): fix issues

arturum1:
- Update macro, file, and signal names to match new ones from build-lib
- Update macros and parameters from iob_soc_setup.py
- Add function calls in iob_soc_setup.py to setup CACHE and UART
- Update submodules
---
Current status:
- Setup of build directory is successful.
- Simulation builds, however need to implement `rvalid` signal for the UART iob native interface


config_setup.mk
```diff
@@ -26,7 +26,7 @@ LIB_DIR ?= $(SOC_DIR)/submodules/LIB
 PERIPHERALS ?=UART
 
 # default configuration
-CONFIG ?= base
+#CONFIG ?= base #Not used?
```
jjts, line 29: Delete

hardware/simulation/system_top.vt
```diff
@@ -76,11 +76,11 @@ module system_top
      #(
  `ifdef DDR_INIT
        .FILE("firmware.hex"),
-       .FILE_SIZE(`FW_SIZE),
+       .FILE_SIZE(`IOB_FW_SIZE),
```
jjts, line 79: IOB_SOC?

PR #488, opened 2022-12-12
Update submodules, pass en_i to peripherals;

PedroAntunes178:


jjts:
In any case, the enable should be a primary input of the system, not generated internally.

PR #494, opened 2022-12-21
*_setup.mk: progress in deprecating setup.mk file tree.

PedroAntunes178:


jjts:
pls use branch python-setup

Issue #529, opened 2023-02-10
DCACHE_ADDR_W to become MEM_ADDR_W


arturum1:
It is a property of the Board or Simulator and should be removed from the parameter list in *_setup.py

PR #530, opened 2023-02-14
Add default.nix; Update README; Resolve issues.

arturum1:
- Remove DCACHE_ADDR_W from iob_soc_setup.py as the new macro is set by the new LIB. This resolves https://github.com/IObundle/iob-soc/issues/529.
- Along with the new LIB, this resolves issue IObundle/iob-lib#430.
- Update README with instructions for usage of Nix environment.
- Add default.nix symlink to new shell.nix of new LIB.
- Update README according to new python-setup process.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,9 +27,9 @@ Older versions of the dependencies above may work but were not tested.
 
 ## Nix environment
 
-Instead of manually installing the dependencies above, you can use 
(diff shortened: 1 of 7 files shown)


jjts:
Looks good, but FPGA checks are failing. Let me know if you need help.

PR #538, opened 2023-03-15
Rename `iob_soc_top.vt` to `iob_soc_sim_wrapper.vt`

arturum1:
- Rename `iob_soc_top.vt` to `iob_soc_sim_wrapper.vt`
- Add mising `iob_lib.vh` include in `iob_soc_tb.vt`
- Update LIB


arturum1:
The aes-ku040 check seems to have failed due to multiple user access to the board.

jjts:
test for kintex board failed because makefile returned status 2

need to check why , it has happened with other makefiles

PR #549, opened 2023-04-26
feat(format): format python, update LIB

P-Miranda:
- update LIB submodule with format-check for python code
- format python code in repository
- format check runs during `setup` - if format-check fails, run `make format` and `make setup` again


P-Miranda:
Checks fail due to `black` missing in target machines

PR #564, opened 2023-06-16
Shared external memory

arturum1:
This PR enables IOb-SoC to share the external memory with its peripherals.
The IOb-SoC splits the memory into equally sized address spaces, one for itself and the others for its peripherals.
The number of address spaces created is the minimum required to give each peripheral its own.
If there are no peripherals that use external memory, then there is only one address spaces that spans the entire memory address space, and is used only for the IOb-SoC system.

The peripherals may provide buses with multiple axi_m interfaces. In that case, each interface of the bus provided by the peripheral will get its own memory address space.

Related to task: https://github.com/orgs/IObundle/projects/1?pane=issue&itemId=30349667


jjts (commented):
This is not the intended functionality. Each thing connected to the external memory should see the whole address space. 

System and peripherals share the memory. They do not divide the memory. For example, Versat is an accelerator that computes arrays for the processor to use. Iob-eth accesses system buffers to send and receive.

arturum1:
Yes, they can all see entire memory, but the starting address is different. 
For example, if a new peripheral is added and requests memory address 0, the IOb-SoC will convert that address to the starting address of the memory zone attributed to it (by inverting the highest bits).
If this peripheral wants to access the IOb-SoCs memory zone, it can by using the higher addresses.

arturum1:
Another example, if the memory has 24 bits (Address space 0x000000 to 0xFFFFFF)
And IOb-SoC has 3 peripherals that use external memory, then the IOb-SoC (using the `iob_addr_zone_selector` Verilog component) will divide the memory into 4 zones:
- 0x000000 to 0x3FFFFF -> Allocated to iob-soc (base address 0x000000)
- 0x400000 to 0x7FFFFF -> Allocated to peripheral 1 (base address 0x400000)
- 0x800000 to 0xBFFFFF -> Allocated to peripheral 2 (base address 0x800000)
- 0xC00000 to 0xFFFFFF -> Allocated to peripheral 3 (base address 0xC00000)

Therefore, the base address of iob-soc is 0x000000, but it can still use any of the other zones by selecting addresses within their range.

When peripheral 1 tries to access address 0x000000, the IOb-SoC system will convert to the real memory address 0x400000 (by inverting bit 22).
Therefore, if the peripheral 1 wants to access the memory zone of IOb-SoC, it should request for addresses starting at memory zone 0x400000. The IOb-SoC will convert this address (by inverting bit 22), to the real memory address 0x000000.

jjts:
This is the Tester/SUT scheme generalized to N memory sharers. It is excellent for having multi-processors in the future that share memory.

However, it is impractical for each CPU+peripherals group because it requires frequent address translations.

From the programmer's point of view, the CPU produces an array A. Now it wants to tell peripheral X that the array is in position &A. Peripheral X will take A and compute array B, for it has special hardware that can do this much faster than the CPU. The addresses &A and &B are stored in two peripheral registers.

However, peripheral X does not see A and B in the same address as the CPU. Address translation is necessary. Is this done in hardware or software? Which item needs to worry about this? The CPU or the peripheral?

The current scheme is acceptable when CPUs share data because they do so less frequently. But this is only easy for peripherals if we develop a way to make it transparent for the programmer.

PR #574, opened 2023-07-18
Update scripts to use `iob_module` instances instead of the old `iob_verilog_instance` class.

arturum1:
- Update scripts according to changes in PR https://github.com/IObundle/iob-lib/pull/625


jjts (commented):
Cant merge due to conflicts

PR #592, opened 2023-08-25
[general] added cke_i to iob_soc

AndreMerendeira:


scripts/iob_soc_create_wrapper_files.py
```diff
@@ -94,8 +94,8 @@ def create_interconnect_instance(out_dir, name, num_extmem_connections):
       .S_COUNT     ({num_extmem_connections}),
       .M_COUNT     (1)
    ) system_axi_interconnect (
-      .clk(clk),
-      .rst(rst),
+      .clk(clk_i),
+      .rst(arst_i),
```
arturum1, line 98: These lines are used to generate the interconnect instance in every wrapper that needs one.
Therefore, the signals generated must have a common name to all wrappers.

Currently, the `iob_soc_sim_wrapper.v` and the `iob_soc_fpga_wrapper.v` of the CYCLONEV are the only wrappers that use this interconnect, and they contain the common wires signals `clk` and `rst`.
They do NOT contain the `arst_i` signal, therefore trying to generate an interconnect with this signal will give an error (that is why those checks are failing).

I suggest keeping the common `clk` and `rst` signals.
```suggestion
      .clk(clk),
      .rst(rst),
```

Another solution would be to make sure the `clk_i` and `arst_i` signals exist in the wrappers.

arturum1 (changes requested):

jjts:
Rst cannot connect to arst

They have a different meaning

Use arst for async reset (posedge)

Use rst for sync reset (no posedge)

jjts:
Checks fail , can't merge

jjts:
the author has no time to fix this - closing

Issue #618, opened 2023-11-02
mkregs address macros same name as constants


jjts:
For example, IOB_X_VERSION is a wrong address, and the value of its constant.

The address should be called IOB_X_VERSION_ADDR

PR #619, opened 2023-11-03
fix(SWREG): <SWREG>_ADDR as macro address; remove unused <SWREG>_rvalikd_rd signal

P-Miranda:
- update software macro addresses for software accessible registers to
  have `_ADDR` sufix
- remove unused SWREG_rvalid_rd signal. Now swreg module generates
  rvalid signal for all software accessible registers
- Addresses #618


P-Miranda:
cyclonev test failed due to some console conflict
rerun the job should pass

Issue #632, opened 2023-11-09
delete pre-setup /if_gen


jjts:
# override empty iob_module's method
    @classmethod
    def _pre_setup(cls):
        # Pre-setup specialized IOb-SoC functions
        pre_setup_iob_soc(cls)

jjts:
Duplicate of #606

Issue #629, opened 2023-11-09
incomprehensible code snippet /if_gen


jjts:
This must go

# Remove [0+:1] part select in AXI connections of ext_mem0 in iob_soc.v template
        if cls.num_extmem_connections == 1:
            inplace_change(
                os.path.join(cls.build_dir, "hardware/src", cls.name + ".v"),
                "[0+:1])",
                ")",
            )

jjts:
stale or done?

arturum1:
Resolved. This snippet was removed in commit https://github.com/IObundle/iob-soc/commit/957b584c0e7ff6333395f128387fef8a6a320733#diff-7c0ac01ecd5af35a96dd499e6888f13b879c4fb981317a5eca7a794945ea6ee1L247-L257.

Issue #606, opened 2023-10-19
user unfriendly method /if_gen


jjts:
https://github.com/IObundle/iob-soc/blob/d25bc62e6c671918e4410b9e041ab62c64e89993/iob_soc.py#L408

this method is unfriendly: it is too much of burden to understand what can be done with it 

Please remove this method

arturum1:
The `_pre_setup()` method no longer exists in branch if_gen2

jjts:
only main counts

PR #634, opened 2023-11-09
Resolve issues

### LINKED ISSUE #605: do not update define /if_gen
https://github.com/IObundle/iob-soc/blob/f5bbbbc674d9d134cc5a9d9756ae408b3b19777c/iob_soc.py#L351C5-L351C5

check argv before assigning defines

not sure we need update_define() function

### LINKED ISSUE #630: no interfaces in submodules list /if_gen
(iob_ram_sp, {"purpose": "fpga"}),
            {
                "interface": "axi",
                "file_prefix": "",
                "wire_prefix": "",
                "port_prefix": "",
                "widths": {
                    "ID_W": "AXI_ID_W",
                    "ADDR_W": "AXI_ADDR_W",
                    "DATA_W": "AXI_DATA_W",
                    "LEN_W": "AXI_LEN_W",
                },
            },
            {
                "interface": "axi",
                "file

...

do not mix interfaces and submodules

### LINKED ISSUE #631: unneeded code /if_gen
# Get number of peripherals and IDs
        # TODO revise name of function
        get_peripheral_macros(cls.confs, cls.peripherals)

arturum1:
- Add comments to scripts related to issues:
  - https://github.com/IObundle/iob-soc/issues/629.
  - https://github.com/IObundle/iob-soc/issues/606.
  - https://github.com/IObundle/iob-soc/issues/632.
- Check argv when assigning defines. Resolves #605.
- Add dedicated interfaces list. Resolves #630.
- Rename get_peripheral_macros function. Resolves #631.
- Move the `submodule_utils.py` file from the LIB to the `scripts/` directory of the IOb-SoC repository [b43a42f].
- Add a new `init_attr` method that allows us to initialize a class attribute with a default value, if it has not been initialized previously.

### DIFF
--- a/iob_soc.py
+++ b/iob_soc.py
@@ -12,7 +12,7 @@ if __name__ == "__main__":
 
 from iob_block_group import iob_block_group
 from iob_soc_utils import pre_setup_iob_soc, post_setup_iob_soc
-from submodule_utils import append_peripheral_macros
+from submodule_utils import setup_peripherals
 from verilog_tools import inplace_change
 
 # Submodules
@@ -437,8 +437,7 @@ class iob_soc(iob_module):
         ]
 
         # Get number of peripherals and IDs
-        # TODO revise name of function
-        append_peripheral_macros(cls.confs, cls.peripherals)
+        setup_peripherals(cls.confs, cls.peripherals)
 
     # override empty iob_module's method
     @classmethod
--- a/iob_soc.py
+++ b/iob_soc.py
@@ -12,8 +12,7 @@ if __name__ == "__main__":
 
 from iob_block_group import iob_block_group
 from iob_soc_utils import pre_setup_iob_soc, post_setup_iob_soc
-from mk_configuration import update_define
-from submodule_utils import get_peripheral_macros
+from submodule_utils import append_peripheral_macros
 from verilog_tools import inplace_change
 
 # Submodules
@@ -439,7 +438,7 @@ class iob_soc(iob_module):
 
         # Get number of peripherals and IDs
         # TODO revise name of function
-        get_peripheral_macros(cls.confs, cls.peripherals)
+        append_peripheral_macros(cls.confs, cls.peripherals)
 
     # override empty iob_module's method
     @classmethod
--- a/iob_soc.py
+++ b/iob_soc.py
@@ -61,6 +61,8 @@ class iob_soc(iob_module):
 
         """Setup this system using specialized iob-soc functions"""
 
+        # *** SCRIPT BELOW WILL BE REMOVED IN `python-gen` *** https://github.com/IObundle/iob-soc/issues/629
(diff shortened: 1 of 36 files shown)


iob_soc.py
```diff
@@ -70,10 +69,10 @@ def _init_attributes(cls):
             )
 
         # warning: do not initialize lists as class variables, they will be shared between all subclasses
-        cls.peripherals = []
-        cls.peripheral_portmap = []
+        cls.init_attr("peripherals", [])
```
jjts, line 72: I don't get this. If this attribute exists why do I have to initialize it?

arturum1: Because the attribute may not exist.
It only exits if there is a subclass of IOb-SoC that defines it before. Otherwise, it will not exist in IOb-SoC.

arturum1: Note that the `peripherals` list is not an attribute of iob_module. So it must be defined in either the iob-soc class or one of its subclasses.
If the iob-soc does not have subclasses, then it must be defined in the iob-soc class. That is why line 72 is there, to ensure that it is defined.

jjts: There is no SoC without at least one peripheral 

why not 

cls.peripherals = [uart]


?

jjts: why         cls.init_attr("peripherals", []) ?

        cls.peripherals = [] seems better

arturum1: Because if we do have a subclass, for example opencryptolinux, that subclass may want to initialize its own peripherals list first and then call the `super()._init_attributes` to add the iob-soc peripherals.
So, if the iob-soc has this line `cls.peripherals = []`, then it will override any `cls.peripherals` list that may have been defined previously (by the subclasses).
Example:
```
class iob_soc_opencryptolinux(iob_soc):
    ...
    @classmethod
    def _init_attributes(cls):
        ...
        cls.peripherals = [iob_spi_master]  # Add iob_spi_master peripheral
        super()._init_attributes(cls)  # Call _init_attributes of iob-soc
    ...
```
```
class iob_soc(iob_module):
    ...
    @classmethod
    def _init_attributes(cls):
        ...
        cls.peripherals = []  # Will override any `cls.peripherals` defined previously (like the one from opencryptolinux)
```

arturum1: > There is no SoC without at least one peripheral
> 
> why not
> 
> cls.peripherals = [uart]
> 
> ?

For a similar reason to my previous comment.
If we have a subclass of iob-soc (like opencryptolinux) that has previously defined a list of peripherals, this iob-soc line would override that list.

The current line that appends the UART to the peripherals list is this one: https://github.com/IObundle/iob-soc/pull/634/files#diff-551ef7ec958375dc7823b60fea7a6b8ff010a80b79c53afdd937d66369e41807R218
This line does not override the existing peripherals list, it only appends a UART instance at the end of it.

PR #659, opened 2023-12-10
caravel

Vasco-Luz:
Added caravel submdule

added a line to iobsoc5 top makefile to build caravel

added a python script to do osme modifications to the caravel makefile


jjts:
This PR should go to branch caravel.

PR #660, opened 2023-12-10
caravel

Vasco-Luz:
makefile changed

added a python script

added some verilog files for test


.gitmodules
```diff
@@ -7,3 +7,6 @@
 [submodule "submodules/AXI"]
 	path = submodules/AXI
 	url = git@github.com:IObundle/verilog-axi.git
+[submodule "submodules/caravel_user_project"]
```
jjts, line 10: Please use IObundle's fork https://github.com/IObundle/caravel_project as discussed

Makefile
```diff
@@ -71,3 +72,23 @@ board_server_status:
 	systemctl status board_server
 
 .PHONY: setup sim-test fpga-test doc-test test-all board_server_install board_server_uninstall board_server_status
+
```
jjts, line 75: This Makefile does not need to be changed as we will not use it for now

hardware/caravel/src/iob_counter.v
```diff
@@ -0,0 +1,37 @@
+`timescale 1ns / 1ps
```
jjts, line 1: root directory cannot contain scripts -- use the scripts directory

Makefile
jjts: It is used only to call targets in the build directory, which can be called there before things are stable.

submodules/LIB/build.mk
```diff
@@ -113,6 +113,18 @@ fpga-debug:
...
 
+
+
+
+#
+#Caravel
+#
+Caravel_DIR=~/iob_soc/submodules/caravel_user_project 
+caravel-build:
+	make -C $(Caravel_DIR) -j1 build
```
jjts, line 124: this file becomes the Makefile at the root build directory 

CARAVEL must be inside the build directory , not in the repo space 

Use relative paths -- no need for absolute paths 

make -C caravel is enough if CARAVEL_DIR is at the top of the build dir

submodules/LIB/scripts/caravel_gds.py
```diff
@@ -0,0 +1,8 @@
+import os
```
jjts, line 1: This seems to be a build script -- use only one build script and one setup script

jjts (commented):
10 files in this PR is too much since you do not understand the process yet.

Please close this PR and resubmit small 1-3 files PRs

PR #661, opened 2023-12-10
added caravel_user_project as submodule

Vasco-Luz:
added the https://github.com/efabless/caravel_user_project as a submodule for iobsoc.
caravel_user_project is conjunction of scripts that build the open tools and pdks to build a user chip based on verilog/analog. The process of caravel is automatized for digital design.
The avalable PDKS are sky130 node A and node B, and Gfmcu180 node D.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -7,3 +7,6 @@
 [submodule "submodules/AXI"]
 	path = submodules/AXI
 	url = git@github.com:IObundle/verilog-axi.git
+[submodule "submodules/caravel_user_project"]
+	path = submodules/caravel_user_project
+	url = https://github.com/efabless/caravel_user_project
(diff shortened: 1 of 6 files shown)


.gitmodules
```diff
@@ -7,3 +7,6 @@
 [submodule "submodules/AXI"]
 	path = submodules/AXI
 	url = git@github.com:IObundle/verilog-axi.git
+[submodule "submodules/caravel_user_project"]
```
Vasco-Luz, line 10: The addition of the submodule to iob-soc

Makefile
```diff
@@ -71,3 +71,18 @@ board_server_status:
 	systemctl status board_server
 
 .PHONY: setup sim-test fpga-test doc-test test-all board_server_install board_server_uninstall board_server_status
+
+
+
+
+
```
jjts, line 78: not needed

hardware/caravel/src/iob_reg.v
```diff
@@ -0,0 +1,22 @@
+`timescale 1ns / 1ps
```
jjts, line 1: not needed

jjts:
This PR is only slightly better than the previous one. Closing.

PR #663, opened 2023-12-10
added submodule

Vasco-Luz:
Added caravel_user_project submodule.
Are a conjuction of scripts that install the pdk and the requirement tools for the design of the caravel chip using sky130 or gf180


jjts:
wrong branch

PR #678, opened 2023-12-13
fix(fpga): cyclonev fpga wrapper, reset_sync

P-Miranda:
- only declare clk_interconnect and arst_interconnect signals if Cyclonev is using external memory
- fix reset sync module: 
  - declare sync as `wire` instead of `reg` 
  - instantiate `iob_r` with correct `RST_VAL` in each `RST_POL` case

### DIFF
--- a/submodules/LIB/hardware/modules/sync/iob_reset_sync/hardware/src/iob_reset_sync.v
+++ b/submodules/LIB/hardware/modules/sync/iob_reset_sync/hardware/src/iob_reset_sync.v
@@ -8,29 +8,38 @@ module iob_reset_sync (
 );
 
    wire [1:0] data;
-   reg [1:0] sync;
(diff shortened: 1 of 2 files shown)


submodules/LIB/hardware/modules/sync/iob_reset_sync/hardware/src/iob_reset_sync.v
```diff
@@ -8,29 +8,38 @@ module iob_reset_sync (
 );
 
    wire [1:0] data;
-   reg [1:0] sync;
+   wire [1:0] sync;
 
    localparam RST_POL = `IOB_REG_RST_POL;
```
jjts, line 14: localparam RST_VAL = `IOB_REG_RST_POL? 2'd0: 2'd3;

no generate

PR #692, opened 2023-12-27
[module.py] changed rw_overlap default to False

AndreMerendeira:


jjts:
Why does such a simple change crash everything?

PR #696, opened 2024-01-03
Caravel 2

Vasco-Luz:
modified the python so it doens include certain .v and certain .vh in the json file of the caravel module


jjts:
Please fix the failing checks before I can merge

PR #713, opened 2024-01-13
Update caravel_iob_soc to the newest version of iob_soc_main

Vasco-Luz:
Updating the version of iob_soc_caravel to the same version of iob_soc_main, because the version of iob_soc_caravel was far being in terms of comits


jjts:
This is has conflict on the pICORV32 submodule 

Update it with 

git submodule update --init --recursive 

and push to update this PR

PR #714, opened 2024-01-14
Modified the python script, deleted some unencessary lines in setup.mk modified the top.v that is being used to test

Vasco-Luz:
Baiscly changed some thing in the python script that now it searche the instasiated modules and the path to those modules to the json file. it does a search for instasiated modules in the previus talked modules so the json has the path to all .v files and .vh files.

Deleted some lines that were uncessessary, blank spaces.


jjts:
A clean PR - congratulations!

PR #725, opened 2024-01-21
change, added iobtowishbone module to the iob soc python script, and fixed a bug in the setup caravel scrit, changed the name of the top module 

Vasco-Luz:
In the first part added wishbone to the modules to be setup into the build directory, so a change in the iob_soc.py had to be done.
changed the name of iob_soc_test to iob_soc_caravel because its a better name.
instantiated the iobtowishbone in caravel, but have questions to how to include with iob_soc module
changed something in the python setup script because there was a bug in the case of wishbone since various reg are instatiated in the same module, it ended adding varius time the same file to the json file, that would induce it into a a problem.

### DIFF
--- a/hardware/caravel/src/iob_soc_caravel.v
+++ b/hardware/caravel/src/iob_soc_caravel.v
@@ -106,7 +106,7 @@ module iob_soc_caravel #(
 
 
 //on works
-iob_iob2wishbone #(
+iob_wishbone2iob #(
    parameter ADDR_W = 32,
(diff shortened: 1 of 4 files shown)


Vasco-Luz:
forgot to say, build_caravel was removed from the make setup, for time reasons, now build_caravel will have the addition, that will automaticly install dependencies when runned and will output (or try in this beggining stages) the gds of the module, a process that takes aprox 45min to an 1h

PR #727, opened 2024-01-22
used localparams for constants

AndreMerendeira:


jjts:
looks like iob_module changes are breaking this , is that right ?

submodules/LIB/scripts/iob_module.py
```diff
@@ -430,7 +432,7 @@ def _generate_sw(cls, mkregs_obj, reg_table):
             mkregs_obj.write_swheader(
                 reg_table, cls.build_dir + "/software/src", cls.name
             )
-        mk_conf.conf_h(cls.confs, cls.name, cls.build_dir + "/software/src")
+            mk_conf.conf_h(cls.confs, cls.name, cls.build_dir + "/software/src")
```
arturum1, line 435: > looks like iob_module changes are breaking this , is that right ?

yes, the iob-soc system does not have registers (no `cls.regs` dictionary), therefore, this change is preventing the system from generating the `iob_soc_conf.h` header.
Macros from this header are used in the IOb-SoC firmware.
For example, the peripheral numbers defined in that header are used to calculate their base addresses, with help from the `iob_soc_periphs.h` file.

The Tester system uses even more of these macros, like detecting when the system runs with INIT_MEM=1:
https://github.com/IObundle/iob-soc-sut/blob/1e503de24c18420471cc53887d7140d8499db6d2/submodules/TESTER/software/src/iob_soc_tester_firmware.c#L160
Or calculating memory addresses:
https://github.com/IObundle/iob-soc-sut/blob/1e503de24c18420471cc53887d7140d8499db6d2/submodules/TESTER/software/src/iob_soc_tester_firmware.c#L327

submodules/LIB/scripts/iob_module.py
```diff
@@ -419,8 +421,8 @@ def _generate_hw(cls, mkregs_obj, reg_table):
     @classmethod
     def _generate_sw(cls, mkregs_obj, reg_table):
         """Generate common software files"""
-        os.makedirs(cls.build_dir + "/software/src", exist_ok=True)
         if cls.regs:
+            os.makedirs(cls.build_dir + "/software/src", exist_ok=True)
```
arturum1, line 425:
```suggestion
        os.makedirs(cls.build_dir + "/software/src", exist_ok=True)
        if cls.regs:
```

submodules/LIB/scripts/iob_module.py
arturum1: Therefore I think we should revert the changes from lines 425 and 435 of `iob_module.py`
And this way, keep generating the `*_conf.h` header for systems/cores without registers.

submodules/LIB/scripts/iob_module.py
AndreMerendeira: The problem with that is that every module will create a _conf.h file, adding dozens of unused files

submodules/LIB/scripts/iob_module.py
AndreMerendeira: The problem with that is that every module will create a _conf.h file, adding dozens of unused files

arturum1: Are unused files an issue?
We used to have the `flows` attribute to prevent that, however, we decided to deprecate that attribute and now every 'flow' is set up for every core (therefore, the software files are also copied/generated for every core even if they are not used).
I'm not sure why we decided to remove the flows, but I assume it's because there was no issue with having unused files in the build directory.
The flows were removed on this PR: https://github.com/IObundle/iob-soc/pull/651

arturum1: The old `emb` flow would prevent modules from generating the `*_conf.h` file if they did not have this flow.
https://github.com/IObundle/iob-soc/pull/651/files#diff-352ff413d1b7a3a396110618b053344ae8d4ed65f727a5a38de8505151123578L407-L419

jjts:
The flows have been removed after we could not remember why they existed. Everything that can not be explained to users must be removed...

jjts:
Perhaps we need something like IS_FPGA to tell us we need to create the software.

AndreMerendeira:
Isn't there a is_top_module attribute or something similar?

arturum1:
We do have the `is_top_module` attribute but I don't think that's what we need.
In the Tester+SUT example, the Tester is the top module while the SUT is not, and we need to have both the `iob_soc_sut_conf.h` and `iob_soc_tester_conf.h` in the build directory for both systems to work properly.

AndreMerendeira:
What if we just add the version swreg to iob-soc? Can work as a good workaround

jjts:
I suggest André withdraw the change to iob_module so that this PR can be merged, and Artur fixes the software flow using the new is_system attribute we discussed privately.

PR #728, opened 2024-01-23
feat(is_system): Add `is_system` attribute to `iob_module`

arturum1:
The new `is_system` attribute, when set, causes the copy/generation of software and related files during the setup process.
It also add the `IS_FPGA` makefile variable to the `config_build.mk` file of the build directory.

Related to discussion in https://github.com/IObundle/iob-soc/pull/727


jjts:
There is more to this.

1) Peripherals also need to copy their sw drivers to the build directory, and the firmware under the TESTER directory if available. This is because whoever receives a peripheral will get an example firmware.

2) The _conf.h files should not be created for modules which have the confs attribute empty.

3) The VERSION macro is only added if the regs attribute is non empty -- it should not be added to all modules.

PR #733, opened 2024-01-25
[mkregs] fixed double rvalid

AndreMerendeira:


jjts:
This is not the path we want for our simple iob interface ....

PR #746, opened 2024-02-06
feat(iob_timer): add Timer linux drivers and user program example

P-Miranda:
- add iob_timer linux drivers
- iob_timer user program example
- update LIB python scripts to support linux header generation from iob_module.py:
```bash
# example for iob_timer linux header generation
./path/to/bootstrap.py iob_timer -f gen_linux_driver_header -o path/to/output_dir
```
  - required some refactoring in `mkregs.py` since linux header has multiple defines in common with `iob_module_swreg.h`
- add README with file information
- add device tree template to manually add iob_timer node to system device tree to enable hardware detection by the linux kernel


jjts:
does it break I2S?

AndreMerendeira (approved):
These changes will probably not break png or i2s

jjts:
please learn to check out PRs and try  -it takes seconds to do 

probably is not good enough

arturum1:
I2S: I have updated the IOBSOC submodule in the i2s repo to use this branch, and the target `make test` still works.
PNG: Cannot test on png repo because main branch does no pass checks (even without trying to update IOBSOC).

arturum1 (changes requested):

PR #750, opened 2024-02-09
Added a iob_soc_wrapper.v inside common_src added a ram to the top

Vasco-Luz:
In the iob_soc_wrapper.v the iob_soc is instatiated, and the rams iob_ram_sp are instatiated. the source file of the iob_soc and the other modules havent suffer any change in terms of ports and wires. its just to check if the procedure of the instatiation of the rams in the top level is being done acordingly.


Vasco-Luz:
The wires on the iob top were just to remember the dimension of the wire. a real rom implementation was added

PR #752, opened 2024-02-10
added the src_common, added a new memories

Vasco-Luz:
Memories doesnt fully work as suposed. but all ram rom .v are copied to a new directory in build/hardware/memories


Vasco-Luz:
instantiated the iob_soc in the wrapper, and made so the module to be simulated is the wrapper and not iob_soc, going to see if it passes the testes

Issue #759, opened 2024-02-15
create_wrapper_files


jjts:
What wrapper what files? Isn't there a better name?

cyclonev related scripts should go to the cyclonev directory

do not shebang files that do not run standalone and do not give them x permission (fixed)

jjts:
Add a header comment explaining what the files does

arturum1:
For now, every iob-soc python script (except for the main iob_soc.py) is located in the `scripts/` directory.
Should I create new scripts in each of the boards/simulation directories with functions dedicated for them?

jjts:
yes, please

each board, simulator, etc has its own folder 

except synthesis as for now we only use one tool

jjts:
some  files belong to the FPGA compile tool, not to any specific board

arturum1:
Removed create_wrapper_files script in commit https://github.com/IObundle/iob-soc/pull/906/commits/834f533cad7e97e739bd43204514116c24f95cff.
Currenlty, in if_gen2, fpga and simulation wrappers are generated with py2hwsw.

PR #767, opened 2024-02-16
Add header comment to `iob_soc_create_wrapper_files.py`

arturum1:
Add comment header to `iob_soc_create_wrapper_files.py` as suggested in https://github.com/IObundle/iob-soc/issues/759


jjts:
pls rename file to iob_soc_create_vs_snippets

arturum1:
I will close this for now as I still need to do the following:
1) Add `board_setup.py` for each board (and simulation?)
2) Update `iob_module.py` to call these scripts based on the `board_list` attribute.

PR #770, opened 2024-02-16
removed the sram reg defenition and switched to iob_reg

Vasco-Luz:
Removed reg defenition and added iob_reg, there is a problem still.

when rom reg signals are switched to registers the iob_soc stops working still tryign to understand.


hardware/src/iob_soc_boot_ctr.v
```diff
@@ -111,25 +126,80 @@ module iob_soc_boot_ctr #(
...
+   
+
+   iob_reg_r #(
+      .DATA_W (1),
+      .RST_VAL(1)
+   ) sram_w_valid_reg (
+      .clk_i (clk_i),
+      .arst_i(arst_i),
+      .cke_i (cke_i),
+      .rst_i (1'b0),
```
jjts, line 152: rst_i is not used then; use iob_reg instead

PR #771, opened 2024-02-16
reset of the error

Vasco-Luz:
made an bad error, this is a pr to revert the error

### DIFF
--- a/hardware/src/iob_soc_boot_ctr.v
+++ b/hardware/src/iob_soc_boot_ctr.v
@@ -132,8 +132,12 @@ module iob_soc_boot_ctr #(
 
    assign sram_w_valid_nxt = boot_o ? rom_r_valid : 1'b0;
 
-   reg [SRAM_ADDR_W-2-1:0] sram_w_addr; 
 
+
+   wire [SRAM_ADDR_W-2-1:0] sram_w_addr;
+   wire [SRAM_ADDR_W-2-1:0] sram_w_addr_nxt;
+
+   assign sram_w_addr_nxt = boot_o ? {3'b111, {(BOOTROM_ADDR_W - 2) {1'b0}}} + rom_r_addr : {3'b111, {(BOOTROM_ADDR_W - 2) {1'b0}}};
 
 
    iob_reg_r #(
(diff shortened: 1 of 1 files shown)


jjts:
This PR isn't necessary , just the one with the iob_reg instead of always

Vasco-Luz:
an error is happening when .rst_i (boot_0) the system finds cpu trap condition

Vasco-Luz:
solved,  syncronous reset cannot be dependent of boot_o because its because of boot_o that the output of sram_valid it becames equal to rom_valid. so it a 0 is putted in rst_i because there is no need for synchronous reset.

Vasco-Luz:
{3'b111, {(BOOTROM_ADDR_W - 2) {1'b0}}}
this line must substitute
{1'b1, {(BOOTROM_ADDR_W - 2) {1'b0}}}
in order for the sram pointer to point to the correct memory location.

Vasco-Luz:
changes, iob_reg_r to iob_reg, ((2**SRAM_ADDR_W-2**BOOTROM_ADDR_W)>>2) to get the righ pointer start(bits that point to the ram memory), changed so the reset of the wstrb is actually a 0.

Vasco-Luz:
(1'b1<<(SRAM_ADDR_W-2))-(1'b1<<(BOOTROM_ADDR_W-2))
changed the way it does the calcuation. to solve syntethesis errors.

hardware/src/iob_soc_boot_ctr.v
```diff
@@ -130,70 +130,73 @@ module iob_soc_boot_ctr #(
...
       .DATA_W (1),
-      .RST_VAL(1)
-   ) sram_w_valid_reg (
+      .RST_VAL(1'b0)
+   )sram_w_valid_reg (
       .clk_i (clk_i),
       .arst_i(arst_i),
-      .cke_i (cke_i),
-      .rst_i (1'b0),
+      .cke_i (1'b1),
```
jjts, line 155: .cke_i(cke_i)

cke_i must come all the way from the top

Vasco-Luz:
removed merge conflitcts with main

PR #793, opened 2024-02-26
update mem_wrap

Vasco-Luz:
Changed the suffix names of the ios of the modules to be easier to read with the suffixx _i or _o in relation if its an input or output.
deleted commented code.


jjts:
The request was to add suffixes to ports but they have been added to wires

PR #794, opened 2024-02-27
Continue devel of `*_gen` files and functions.

arturum1:


jjts:
Why do you call super class methods?

Did you not inherit them?

jjts:
We agreed to use the constructor, not _setup....

arturum1:
> Why do you call super class methods?
> 
> Did you not inherit them?

Yes, the methods were inherited, but if we do not call the ones from the super class, then we would only be overriding the methods. Not extending their functionality.

arturum1:
> We agreed to use the constructor, not _setup....

Yes, for now I am trying to make the system work as is, and afterwards perform the merge between _setup and __init__.
I can open an issue with  this so that we remember to do it.

PR #798, opened 2024-02-29
Sketch for structure with dataclasses and merge `_setup()` into `__init__()`.

arturum1:
This is a draft PR discuss implementation details about merging the `_setup()` method into the `__init__()` method and about the use of python dataclasses.
Related to discussion in https://github.com/IObundle/iob-py2hw/pull/42

Initially I only modified the `iob_soc.py` and `iob_module.py` files.

### DIFF
--- a/iob_soc.py
+++ b/iob_soc.py
@@ -1,11 +1,11 @@
 #!/usr/bin/env python3
 
 import sys
-from dataclasses import dataclass, field
+from dataclasses import dataclass
 
 from iob_module import iob_module
 from iob_block_group import iob_block_group
-from iob_soc_utils import pre_setup_iob_soc, post_setup_iob_soc, find_dict_in_list
+from iob_soc_utils import pre_setup_iob_soc, post_setup_iob_soc
 
 # Submodules
 from iob_picorv32 import iob_picorv32
--- a/iob_soc.py
+++ b/iob_soc.py
@@ -256,66 +256,64 @@ class iob_soc(iob_module):
         uart.instance("UART0"),
         timer.instance("TIMER0"),
     ]  # List of peripherals
-    peripheral_portmap = (
-        [  # List of tuples, each tuple corresponds to a port map
-            (
-                {
-                    "corename": "UART0",
-                    "if_name": "rs232",
-                    "port": "txd",
-                    "bits": [],
-                },
-                {
-                    "corename": "external",
-                    "if_name": "uart",
-                    "port": "uart_txd_o",
-                    "bits": [],
-                },
-            ),
-            (
-                {
-                    "corename": "UART0",
-                    "if_name": "rs232",
-                    "port": "rxd",
-                    "bits": [],
-                },
-                {
-                    "corename": "external",
-                    "if_name": "uart",
-                    "port": "uart_rxd_i",
-                    "bits": [],
-                },
-            ),
-            (
-                {
-                    "corename": "UART0",
-                    "if_name": "rs232",
-                    "port": "cts",
-                    "bits": [],
-                },
-                {
-                    "corename": "external",
-                    "if_name": "uart",
-                    "port": "uart_cts_i",
-                    "bits": [],
-                },
-            ),
-            (
-                {
-                    "corename": "UART0",
-                    "if_name": "rs232",
-                    "port": "rts",
-                    "bits": [],
-                },
-                {
-                    "corename": "external",
-                    "if_name": "uart",
-                    "port": "uart_rts_o",
-                    "bits": [],
-                },
-            ),
-        ]
-    )
+    peripheral_portmap = [  # List of tuples, each tuple corresponds to a port map
+        (
+            {
+                "corename": "UART0",
+                "if_name": "rs232",
+                "port": "txd",
+                "bits": [],
+            },
+            {
+                "corename": "external",
+                "if_name": "uart",
+                "port": "uart_txd_o",
+                "bits": [],
+            },
+        ),
+        (
+            {
+                "corename": "UART0",
+                "if_name": "rs232",
+                "port": "rxd",
+                "bits": [],
+            },
+            {
+                "corename": "external",
+                "if_name": "uart",
+                "port": "uart_rxd_i",
+                "bits": [],
+            },
+        ),
+        (
+            {
+                "corename": "UART0",
+                "if_name": "rs232",
+                "port": "cts",
+                "bits": [],
+            },
+            {
+                "corename": "external",
+                "if_name": "uart",
+                "port": "uart_cts_i",
+                "bits": [],
+            },
+        ),
+        (
+            {
+                "corename": "UART0",
+                "if_name": "rs232",
+                "port": "rts",
+                "bits": [],
+            },
+            {
+                "corename": "external",
+                "if_name": "uart",
+                "port": "uart_rts_o",
+                "bits": [],
+            },
+        ),
+    ]
     # Number of external memory connections (will be filled automatically)
     num_extmem_connections = -1
     # This is a standard iob_module attribute, but needs to be defined after 'peripherals' because it depends on it
@@ -341,18 +339,17 @@ class iob_soc(iob_module):
         iob_block_group(
             name="peripheral",
             description="Peripheral module",
-            blocks=self.peripherals,
+            blocks=peripherals,
         ),
     ]
 
     def __post_init__(self, *args, is_top=True, **kwargs):
         self.is_top_module = is_top
         self.set_default_build_dir()
-        # TODO: Somehow we have to run pre_setup before the __init__
         # Pre-setup specialized IOb-SoC functions
         pre_setup_iob_soc(self)
-        # Call the superclass _setup # TODO: DELETE THIS. This does not exist anymore. Only leaving here for reference.
-        # super()._setup(*args, is_top=is_top, **kwargs)
+        # Call the superclass setup
+        super().__post_init__(*args, is_top=is_top, **kwargs)
         # Post-setup specialized IOb-SoC functions
         post_setup_iob_soc(self)
 
--- a/iob_soc.py
+++ b/iob_soc.py
@@ -39,14 +39,13 @@ class iob_soc(iob_module):
     version = "V0.70"
     rw_overlap = True
     is_system = True
-    board_list = field(default_factory=lambda: [
-        "CYCLONEV-GT-DK", "AES-KU040-DB-G"])
+    board_list = ["CYCLONEV-GT-DK", "AES-KU040-DB-G"]
     cpu = iob_picorv32()
     uart = iob_uart()
     timer = iob_timer()
     merge = iob_merge()
     split = iob_split()
-    submodule_list = field(default_factory=lambda: [
+    submodule_list = [
         cpu,
         iob_cache(),
         uart,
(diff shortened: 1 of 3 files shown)


AndreMerendeira (approved):

iob_soc.py
```diff
@@ -33,325 +34,325 @@
...
+        ),
+        iob_block_group(
+            name="peripheral",
+            description="Peripheral module",
+            blocks=self.peripherals,
+        ),
+    ]
 
-    def _setup(self, *args, is_top=True, **kwargs):
+    def __post_init__(self, *args, is_top=True, **kwargs):
```
arturum1, line 348: This is the new `__post_init__()` method (it was previously `_setup()`).
It is now responsible for setting up the the core right after initializing the attributes with the implicit `__init__()`.

iob_soc.py
```diff
@@ -33,325 +34,325 @@
...
         self.is_top_module = is_top
         self.set_default_build_dir()
+        # TODO: Somehow we have to run pre_setup before the __init__
         # Pre-setup specialized IOb-SoC functions
         pre_setup_iob_soc(self)
-        # Call the superclass _setup
-        super()._setup(*args, is_top=is_top, **kwargs)
+        # Call the superclass _setup # TODO: DELETE THIS. This does not exist anymore. Only leaving here for reference.
+        # super()._setup(*args, is_top=is_top, **kwargs)
```
arturum1, line 355: I'm not currently fully aware of how the dataclasses interact with subclasses, but I assume that the `__post_init__` method will override the `__post_init__()` from the superclass.
Therefore, I think this line will probably become:
```suggestion
        super().__post_setup__(*args, is_top=is_top, **kwargs)
```

lib/scripts/iob_module.py
```diff
@@ -16,40 +17,40 @@
...
-        self.PURPOSE_DIRS = {
-            "hardware": "hardware/src",
-            "simulation": "hardware/simulation/src",
-            "fpga": "hardware/fpga/src",
-        }
-
-    def _setup(self, is_top=True, purpose="hardware", topdir="."):
+    """Generic class to describe how to generate a base IOb IP core / module"""
+
+    name: str = name or self.__class__.__name__
```
arturum1, line 24: This line throws a syntax error when using dataclasses for two reasons:
1) The keyword `self` is not recognized. Probably can be replaced by just `__class__.__name__`
2) The `name` variable is not recognized because it has not been defined before. This variable was previously an argument o the `__init__` method. Now it can probably be removed.

arturum1 (commented):
Here are a few notes about the initial implementation of the new structure.

lib/scripts/iob_module.py
arturum1: This probably makes it work as intended (by default use the class name unless a subclass specifies a new name):
```suggestion
    name: str = __class__.__name__
```

iob_soc.py
```diff
@@ -33,325 +34,325 @@
 from axi_interconnect import axi_interconnect
 
 
+@dataclass
```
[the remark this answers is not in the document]
jjts: We are going to use it to validate user inputs in the core.py file

iob_soc.py
jjts: Do you mean __post_init__?

jjts: pls try in a simple example , ask chatgpt...

iob_soc.py
```diff
@@ -33,325 +34,325 @@
...
-                "type": "master",
-                "wire_prefix": "",
-                "port_prefix": "",
-                "mult": "",  # Will be filled automatically
-                "widths": {
-                    "ID_W": "AXI_ID_W",
-                    "ADDR_W": "AXI_ADDR_W",
-                    "DATA_W": "AXI_DATA_W",
-                    "LEN_W": "AXI_LEN_W",
+    version = "V0.70"
```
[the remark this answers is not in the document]
jjts: Types are used whenever we want arguments to be automatically checked.

We are not adding types to inflict pain upon ourselves but just to make life easier for users (in case they pass the wrong arguments) and for us who don't have to write meaningless code to check stuff.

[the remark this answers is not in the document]
jjts: before we did not have this feature. Now we have. So we might as well use it!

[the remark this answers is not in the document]
jjts: It may be time to have all those dicts changed into classes!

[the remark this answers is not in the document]
jjts: a class creates a type, which can now be easily checked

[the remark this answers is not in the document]
AndreMerendeira: > It may be time to have all those dicts changed into classes!

I would start by creating the classes one by one so we can discuss them and then apply them and start writing the methods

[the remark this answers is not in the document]
arturum1: The [block_groups](https://github.com/IObundle/iob-soc/pull/798/files#diff-551ef7ec958375dc7823b60fea7a6b8ff010a80b79c53afdd937d66369e41807R322-R346) list is one of those lists that we already modified to use objects instead of dictionaries.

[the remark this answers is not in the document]
jjts: I suggest we simplify it to just blocks, a list of blocks only. 

It is currently used to document top-level blocks. 

The groups were created out of laziness -- so that we don't have to group verilog modules inside a larger verilog module

Ideally, an IP core should have a reasonably low number of top-level blocks (< 10 as a rule of thumb), and only these are documented.

Because we had too many we created block groups to artificially decrease them.

In fact, we will not need blocks , once we create IP cores out of other modules.

My suggestion is to delete block groups and manually write the list in the tex file.

In the future this list is easily extracted from the description of the top-level blocks (blocks below the top-level)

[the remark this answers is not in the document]
jjts: Not simplify! Simply erase.

iob_soc.py
```diff
@@ -33,325 +34,323 @@
...
-                "name": "DATA_W",
-                "type": "F",
-                "val": "32",
-                "min": "1",
-                "max": "32",
-                "descr": "Data bus width",
+    version = "V0.70"
+    rw_overlap = True
+    is_system = True
+    board_list = field(default_factory=lambda: [
```
arturum1, line 42: Turns out python dataclasses [do not accept mutable objects as default values for attributes](https://stackoverflow.com/q/53632152/11442904).
This is because [python stores default objects as class attributes](https://docs.python.org/3/library/dataclasses.html#mutable-default-values).

So we cannot just use `board_list = ["CYCLONEV-GT-DK", "AES-KU040-DB-G"]`, as python throws an error.

To resolve this, I followed the suggestions proposed [here](https://stackoverflow.com/a/75677129), and modified all the list attributes to use the `field` function in commit https://github.com/IObundle/iob-soc/pull/798/commits/0f4d1f611769c118968f00a1e8469155aae24cf5.

arturum1: I also ran this example in the python interperter to verify that python does in fact store default objects as class attributes.
```
>>> class a:
...     def __init__(self,b=[]):
...             b+=[1]
...             print(b)
... 
>>> a()
[1]
<__main__.a object at 0x73792c47c710>
>>> a()
[1, 1]
<__main__.a object at 0x73792c47c750>
>>> a()
[1, 1, 1]
<__main__.a object at 0x73792c47c6d0>
```
In this example, `[]` is the default object for the `b` argument, and since it behaves like a class attribute, it will be shared by all 3 instances of the class `a()` that I created.

jjts: > Turns out python dataclasses [do not accept mutable objects as default values for attributes](https://stackoverflow.com/q/53632152/11442904).
> This is because [python stores default objects as class attributes](https://docs.python.org/3/library/dataclasses.html#mutable-default-values).
> 
> So we cannot just use `board_list = ["CYCLONEV-GT-DK", "AES-KU040-DB-G"]`, as python throws an error.
> 

I thougth this list is always passed to the constructor. Why would we want to write the above?



> To resolve this, I followed the suggestions proposed [here](https://stackoverflow.com/a/75677129), and modified all the list attributes to use the `field` function in commit https://github.com/IObundle/iob-soc/pull/798/commits/0f4d1f611769c118968f00a1e8469155aae24cf5.

AndreMerendeira: Can't we use a tuple?

jjts: There's no way fields and lambdas will be accepted....

lib/scripts/iob_module.py
arturum1: This did not work, the `__class__` variable was not recognized. I had to move this to the `__post_init__()` function [here](https://github.com/IObundle/iob-soc/pull/798/commits/d23aafbef7daf857c44145c6bb83cc3debb99426#diff-7db188bf8175e04672ebfa1ed4a534bcfde732969599f553027f872b0c845911R57).

iob_soc.py
arturum1: > Can't we use a tuple?

Tuples do seem to work as an alternative to creating lists with the default_factory. 
We should note that does not allow us to modify the elements of this attribute in the __post_init__, but we could always just overwrite it with a new tuple if needed.
I've made the change to use tuples in the iob_module attributes [here](https://github.com/arturum1/iob-soc/commit/b2615fa164f8bc6be35d3e04c4c171d5305cfcc4#diff-7db188bf8175e04672ebfa1ed4a534bcfde732969599f553027f872b0c845911R37-R42).

arturum1 (commented):
I've made an experiment to see the behavior of trying to extend a list/tuple in subclasses.
This is useful, for example, when opencryptolinux tries to extend the ios list of iob-soc.

```
from dataclasses import dataclass

@dataclass
class a:
    ios: tuple = (1,2,3)

    def print_ios(self):
        print(self.ios)

@dataclass
class b(a):
    ios: tuple = (4,5,6)

    def __post_init__(self):
        self.ios += super().ios

obj_a = a()
obj_b = b()
obj_a.print_ios()
obj_b.print_ios()
```
Output:
```
$ python3 ./test.py 
(1, 2, 3)
(4, 5, 6, 1, 2, 3)
```

---
If we removed this `__post_init__` function from class b:
```
    def __post_init__(self):
        self.ios += super().ios
```
The tuple would be overriden, instead of extended.
```
$ python3 ./test.py 
(1, 2, 3)
(4, 5, 6)
```
---
Do we accept that we will have to add the line `self.<tuple> += super().<tuple>` to the subclasses that wish to extend lists/tuples from their superclasses?
Is there a better way to do this?

jjts (commented):
There is no point changing many .py now. Change only a few to run 1 or 2 examples.
I've made this comment before. You can run individual lib tests. No ned to run the test target for now. If this is not solid you are always  changing tens of files.

jjts:
With 115 files changed,  nobody can review this. Pls make this a regular PR so  i can merge to declutter. In the meanwhile, the real work that needs to be done is unclear.

jjts:
Please indicate which are the main files to review.

Draft PRs are to present an idea or run checks. I believe the former is the case.

iob_soc.py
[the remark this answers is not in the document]
arturum1: Seems like we are going to use `mypy` https://github.com/IObundle/iob-soc/pull/799
So we no longer need the `@validate_types_class` decorator.

Most of the attributes in the `iob_soc.py` class don't need to define their data types because they are already inherited from iob_module (and their data types are defined there).

iob_soc.py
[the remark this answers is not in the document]
arturum1: Should I delete the `block_groups` list completely, or modify it to be a `blocks` list (without groups)?

> In the future this list is easily extracted from the description of the top-level blocks (blocks below the top-level)

From this phrase it seems that we still have a list describing the blocks. Is this the `blocks` list (without groups)?

PR #800, opened 2024-03-05
changes for SYN

Vasco-Luz:
bascly added a new target, a new variable to the syn. and changed something in iob_soc.py to remove unwanted modules for the synthetisis.


Makefile
```diff
@@ -46,6 +52,9 @@ fpga-test:
 	make clean setup fpga-run BOARD=AES-KU040-DB-G INIT_MEM=1 USE_EXTMEM=0 
 	make clean setup fpga-run BOARD=AES-KU040-DB-G INIT_MEM=0 USE_EXTMEM=1 
 
+syn-run: 
+	make syn-build SYN=1 INIT_MEM=0 USE_EXTMEM=0
```
jjts, line 56: If you are calling the syn makefile you should not need the variable

iob_soc.py
```diff
@@ -374,11 +375,14 @@ def _custom_setup(cls):
         # Add the following arguments:
         # "INIT_MEM": if should setup with init_mem or not
         # "USE_EXTMEM": if should setup with extmem or not
+        # "SYN": if should setyp with syn or not
```
jjts, line 378: What is different sbou the syn setup?

jjts (commented):
Comments done

iob_soc.py
Vasco-Luz: the syn setup is the argument that is passed that is then used during the python setup in order for the python script to understand that its building a syn buildir. then it will not copy certain sources to the buildir.

Makefile
Vasco-Luz: i thought it was to mimick how the fpga and sim-test work

jjts: there's no SIM or FPGA variable....

jjts:
closed as not acted upon

PR #803, opened 2024-03-11
Sketch for: list of objects instead of dictionaries; removal of `peripheral_portmap`; Use `py2hw` syntax to generate `iob_soc.v`.

arturum1:
This is a draft PR to discuss implementation details about:
- Replacing dictionaries from lists (like `ios`, `confs`, `regs`) with objects from classes dedicated to describe them.
- Using [py2hw](https://github.com/IObundle/iob-py2hw) based scripts to manage and connect Verilog modules inside others. This removes the need for the `peripheral_portmap` in iob-soc, since these connections are now done via py2hw scripts.
  The `iob_soc.py` file was updated to use py2hw syntax to generate the entire `iob_soc.v` file.
- Removal of the `iob_block_group` class. The `blocks` list no longer contains groups. Now it only contains a list of Verilog instances (blocks).

The main changes are to the `iob_soc.py` file.
Some files were added based on the ones from the [py2hw](https://github.com/IObundle/iob-py2hw) repo.


iob_soc.py
```diff
@@ -78,272 +81,268 @@ class iob_soc(iob_module):
...
+        txd = iob_wire(name='txd', width=1)
+        rxd = iob_wire(name='rxd', width=1)
+        cts = iob_wire(name='cts', width=1)
+        rts = iob_wire(name='rts', width=1)
+
+        # Connect UART wires
+        self.uart0.ios.txd = txd
+        self.uart0.ios.rxd = rxd
+        self.uart0.ios.cts = cts
+        self.uart0.ios.rts = rts
```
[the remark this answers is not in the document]
arturum1: Always connect ios to intermediate wires, to allow connecting one IO to multiple places

iob_soc.py
```diff
@@ -40,308 +43,606 @@ class iob_soc(iob_module):
...
-            "port_prefix": "",
-            "wire_prefix": "",
-            "descr": "Clock, enable, and reset",
-            "ports": [],
+
+    # Create wire for connection with IOs
+    clk_en_rst = iob_interface("clk_en_rst")
+    clk = clk_en_rst.wires.clk
+    arst = clk_en_rst.wires.arst
+    cke = clk_en_rst.wires.cke
```
arturum1, line 185: We can generate a group of wires and access them individually like this.

iob_soc.py
```diff
@@ -40,308 +43,606 @@ class iob_soc(iob_module):
...
+
+    ios = [
+        iob_interface(
+            name="clk_en_rst",
+            type="slave",
+            port_prefix="",
+            wire_prefix="",
+            descr="Clock, enable, and reset",
+            ports=[],
+            connect_to=clk_en_rst,
```
arturum1, line 216: We can connect an interface to another interface with the `connect_to` attribute.
In this case, we are connecting all 3 ports (IOs) from the `clk_en_rst` interface to the 3 wires of the `clk_en_rst` interface created previously.

iob_soc.py
```diff
@@ -40,308 +43,606 @@ class iob_soc(iob_module):
...
+            port_prefix="",
+            wire_prefix="",
+            descr="iob-soc trap signal",
+            ports=[
+                iob_port(
+                    name="trap",
+                    direction="output",
+                    width=1,
+                    descr="CPU trap signal",
+                    connect_to=trap
```
arturum1, line 230: We can connect a single port to a wire (created previously) using the `connect_to` attribute.

iob_soc.py
arturum1, line 1: The section from [line 290 to line 598](https://github.com/IObundle/iob-soc/blob/86093f8b4d2f73577c9c298b7823d6854de918ee/iob_soc.py#L290-L598) represents the entire `iob_soc.v` module body, using `py2hw` syntax.

iob_soc.py
```diff
@@ -40,308 +43,606 @@ class iob_soc(iob_module):
...
+    #
+
+    boot = iob_wire("boot", width=1)
+    cpu_reset = iob_wire("cpu_reset", width=1)
+
+    #
+    # CPU
+    #
+
+    cpu_i_req = iob_wire("cpu_i_req", width=REQ_W)
```
arturum1, line 305: Currently, the `REQ_W` macro does not exist in python.
We could convert this to a string and let the generator insert it as is in the generated Verilog. This would allow us to keep using the `` `REQ_W`` macro in verilog.

iob_soc.py
```diff
@@ -40,308 +43,606 @@ class iob_soc(iob_module):
...
+    cpu_i_resp = iob_wire("cpu_i_resp", width=RESP_W)
+    cpu_d_req = iob_wire("cpu_d_req", width=REQ_W)
+    cpu_d_resp = iob_wire("cpu_d_resp", width=RESP_W)
+
+    cpu = iob_picorv32(
```
arturum1, line 310: As requested, the new constructors of modules no longer create a "verilog module" or a "verilog instance". Now they create an object that represents a mix between the two.

In this line 310, we are creating a Verilog module and also a Verilog instance, since we are giving it Verilog parameters, and connecting its IOs. We will also instantiate it in the iob_soc.v, by [placing it in the `blocks` list.](https://github.com/IObundle/iob-soc/pull/803/files#diff-551ef7ec958375dc7823b60fea7a6b8ff010a80b79c53afdd937d66369e41807R633).

However, in [lines 601-626](https://github.com/IObundle/iob-soc/pull/803/files#diff-551ef7ec958375dc7823b60fea7a6b8ff010a80b79c53afdd937d66369e41807R601-R626), we are just creating Verilog modules, not Verilog instances. These modules are required because they are used elsewhere in the system (when a submodule needs them but does not include them), or they cannot be directly instantiated as Verilog instances (like verilog headers and software sources).

arturum1 (commented):
The new commit 86093f8b4d2f73577c9c298b7823d6854de918ee resolves most of the comments I mentined previously.
This new commit also modifies the `iob_soc.py` to include most of the `py2hw` code required to generate the full `iob_soc.v` file.
Checkout the new `iob_soc.py` file [here](https://github.com/IObundle/iob-soc/blob/86093f8b4d2f73577c9c298b7823d6854de918ee/iob_soc.py).

jjts:
I added a few comments, but this differs from how we should work. 

Please change ONLY iob_soc.py

The other files are not needed before we discuss iob_soc.py

Closing this PR

iob_soc.py
jjts: we have no interest in REQ_W

iob_soc.py
```diff
@@ -212,6 +177,34 @@ class iob_soc(iob_module):
...
+        },
+        )
+
+    txd = iob_wire(name='txd', width=1)
+    rxd = iob_wire(name='rxd', width=1)
+    cts = iob_wire(name='cts', width=1)
+    rts = iob_wire(name='rts', width=1)
+
     ios = [
         iob_interface(
```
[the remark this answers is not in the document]
jjts: This is what's important .

Can you focus on those a  PR only changing iob_soc.py? I'v requester this a few times . No other files matter.

Preferably a 10-line PR.

jjts:
bla

iob_soc.py
arturum1: Removed `connect_to` attribute/argument in commit 86b8840a9b5e377a9ae2ead2a4d6127883c0a874

iob_soc.py
arturum1: Removed this attribute in commit 86b8840a9b5e377a9ae2ead2a4d6127883c0a874

iob_soc.py
```diff
@@ -212,6 +177,34 @@ class iob_soc(iob_module):
...
+        mult="",  # Will be filled automatically
+        widths={
+            "ID_W": "AXI_ID_W",
+            "ADDR_W": "AXI_ADDR_W",
+            "DATA_W": "AXI_DATA_W",
+            "LEN_W": "AXI_LEN_W",
+        },
+        )
+
+    txd = iob_wire(name='txd', width=1)
```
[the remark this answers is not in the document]
arturum1: Resolved in commit 86b8840a9b5e377a9ae2ead2a4d6127883c0a874

iob_soc.py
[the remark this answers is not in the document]
arturum1: > this ahs already been done above , no need to repeat
> 
> I suggest we have class io and list io_list
> 
> clkrsten_io = io("clk_en_rst")
> 
> this creates a set of ios
> 
> then
> 
> io_list += clkrsten_io
> 
> Or you can create a method clkrsten_io = create_io("clk_en_rst")
> 
> than returns the object and addsit t the io_list

Is the `io_list`, mentioned in this comment, equivalent to the `ios` list? (should I rename `ios` to `io_list`?)

Should the new `ios` list be a list of ports only?
Or should it keep being a list of 'interfaces'/'groups' of ports?

iob_soc.py
```diff
@@ -264,85 +260,391 @@ class iob_soc(iob_module):
...
+    #
+
+    boot = iob_wire("boot", width=1)
+    cpu_reset = iob_wire("cpu_reset", width=1)
+
+    #
+    # CPU
+    #
+
+    cpu_i_req = iob_wire("cpu_i_req", width=REQ_W)
```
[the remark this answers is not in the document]
arturum1: I have grouped the `req` and `resp` singals into a single bus in commit 5521b644f42ece38bf44660b849bea4f842d8cc2.
The new bus is called `cpu_i_bus`.

iob_soc.py
```diff
@@ -212,6 +177,34 @@ class iob_soc(iob_module):
             descr="Offset of memory address",
         ),
     ]
+
+    # Create wire for connection with IOs
+    clk_en_rst = iob_interface("clk_en_rst")
```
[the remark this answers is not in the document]
arturum1: Removed intermediate wires between IOs and modules in commit 86b8840

iob_soc.py
arturum1: Currently, this PR only shows a translation from the current `iob_soc.v` implemented in Verilog, to the `py2hw` syntax in Python.
Removing the `REQ_W`, `RESP_W`, and associated macros would probably mean a major restructure of the current iob_soc module implementation.
If the objective is to keep this PR simple to understand, then this change should probably be done in another PR.

arturum1:
I marked this PR ready for review so that it can be merged before it grows too much.

If there is still some aspect of this PR that is not as intended, I can still add commits to correct it.

iob_soc.py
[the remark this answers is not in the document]
arturum1: Resolved in commit 5164a57

Issue #804, opened 2024-03-13
should be iob_interface like the others


jjts:
https://github.com/IObundle/iob-soc/blob/cd83a441a2acf143f21e3c9c471b74e1a729f4a9/iob_soc.py#L183

jjts:
this seems done or dated

arturum1:
No longer applicable.
Interfaces are now defined using the [interface](https://github.com/IObundle/iob-soc/blob/if_gen2/iob_soc.py#L171-L174) attribute of the wires/ports.

PR #806, opened 2024-03-13
refactor(py2hw): Rename wires, ports, and groups

arturum1:
- 'iob_wire' renamed to 'wire'
- 'iob_port' renamed to 'port'
- 'interface' was split into 2 groups: 'wire_group' and 'port_group'
  This resolves https://github.com/IObundle/iob-soc/issues/804
- 'ios' list renamed to 'port_group_list'


jjts:
In a module you first declare the wires and then the instances.

YOu pass the wire for each port.

The port classes should not be derived from wire. They only need two attributes 

connected_to
direction


when you write the Verilog the name of the port is the wire name + the suffix _i or _o according to direction

So, ports don't even need a name. They have nothing to do with wires.

PR #807, opened 2024-03-14
Pr of the synthetesis, stil lack some feutures, will be added.

Vasco-Luz:
It still lacks the self-purpose of the cache, that will be added in this pr, a sdc file was added for syn, in this point the sources are handpicked. two reg defenition were removed from iob_split and from iob_merge, that were creating problems.


jjts:
closed by inaction

PR #813, opened 2024-03-18
refactor(iob_soc.py): Use `create_*` methods to generate wires, ports, and groups.

arturum1:
This PR is a copy of https://github.com/arturum1/iob-soc/pull/4, but this one is to the upstream repository.

As discussed privately, the wires and ports are now generated by using dedicated `create_wire` and `create_port` methods.
- The `create_wire` method creates a wire, and appends it to the local list of wires named `wire_group_list`. (If the wire does not have a group, one will be created automatically.)
  (I also made groups for wires to be consistent with the ports, and because they are useful to make grouped connections of related wires, like the ones from rs232. With groups, we can connect the entire rs232 group to the uart in one line, instead of connecting each wire, one by one.)
- The `create_port` method creates a port, and appends it to the local module's port list named `port_group_list`. (If the port does not have a group, one will be created automatically.)

I have also created similar functions for grouping wires and ports, named `create_wire_group` and `create_port_group`.
- The `create_wire_group` method creates a group of wires and appends it to the list of wires named `wire_group_list`.
- The `create_port_group` method creates a group of ports and appends it to the module's port list named `port_group_list`.

These grouping functions, receive an argument named `wires` (or `ports`), to allow the user to make a custom (non standard) group. I am not sure if this is the best approach, or if we should always force the user to use the `create_wire`/`create_port` methods.
If this argument is not provided or is empty, the function will create the wires/ports automatically, based on the group name.
Example: The "rs232" group name will create the wires/groups of that standard interface automatically.

An alternative way to assign a wire to a group is by using the argument `group` of the `create_wire`/`create_port` methods.

~~To connect a port to a wire/wire_group, use the `connect` argument of the `create_port`/`create_port_group` methods.
This argument receives a string with the name of the wire/wire_group to connect to.~~
Now, when calling the `create_port`/`create_port_group` methods, the corresponding wires will be created and connected automatically.

When instantianting a module, to connect its ports/port_groups to the wires/wire_groups, use the `connect` argument.
~~This argument is similar to the one used in the `create_port`/`create_port_group`.~~

This commit is a work in progress!
- I am not sure if the groups are implemented in a way that is intuitive enough for the user.
- We should discuss how to extract individual wires from groups. For now I have used the `get_wire_from_group(<group_name>, <wire_name>)` function
- We should discuss what datatypes the `connect` argument should receive. For example, if it receives a string, should it assume it is a wire name or a wire_group name?
- ~~We should also discuss if wires and wire_groups should have unique names, since they are global (accessible by any component in the entire project).~~
  Does a wire that is in a group need to have a unique name? Does it need to be different from a wire that is in a different group?

In the `iob_soc.py` file, I have only changed the port section and the start of the iob_soc module description (the part where we instance the cpu). The rest of the module is unchanged keep this commit simple.

Edit: Removed references about 'global' wires.


iob_soc.py
```diff
@@ -177,88 +177,74 @@ def __init__(self, *args, **kwargs):
...
-        cke = clk_en_rst.wires.cke
 
-        trap = port(
+        # This function creates a wire named 'trap' and adds it to a project global list of wires.
+        create_wire(
             name="trap",
+            width=1,
+        )
+        # This function creates a port in the local module's `port_group_list`, and assigns a group to it. It also connects the port to a wire from the global list.
+        create_port(
```
jjts, line 203: Creating a port automatically creates a wire since port is subclass of wire.

Don't see need to create a wire and a port.

iob_soc.py
```diff
@@ -302,19 +301,23 @@ def __init__(self, *args, **kwargs):
...
                 "USE_EXTMEM": USE_EXTMEM
             },
-            io_connections={
-                "clk_i": clk,
-                "arst_i": cpu_reset,
-                "cke_i": cke,
-                "boot_i": boot,
-                "trap_o": trap,
+            connect={
+                "clk_i": get_wire_from_group("clk_en_rst", "clk"),
```
jjts, line 305: No explocit suffix.  _I and _o managed automatically.

jjts (changes requested):
I missed a module instantiation.maybe only in the next pr.

Ports are also wires. Creating a port creates a wire. Don't do both.

Create only internal wires.

iob_soc.py
arturum1: Should we remove the `connect` attribute from the `create_port` and `create_port_group` functions?
If they already create the corresponding port wire automatically, then I think we don't need to specify the wire that connects to it.

arturum1: Removed wires for ports in commit [0cfa061](https://github.com/IObundle/iob-soc/pull/813/commits/0cfa0611c2abe5bf1551a0ea8da33f5b93238487)

iob_soc.py
arturum1: Removed explicit port suffixes in commit [b3765b7](https://github.com/IObundle/iob-soc/pull/813/commits/b3765b71e2cfe4540777af2e32e0ee167e2763a9)

P-Miranda:
> - I am not sure if the groups are implemented in a way that is intuitive enough
  for the user.
  - from my understanding, groups are used for standard interfaces: axi4, iob native, etc
  - what other situations do we use groups?
    - do we create custom groups in other situations?
> - We should discuss how to extract individual wires from groups. For now I have
  used the `get_wire_from_group(<group_name>, <wire_name>)` function

> - We should also discuss if wires and wire_groups should have unique names,
  since they are global (accessible by any component in the entire project).
    - Does a wire that is in a group need to have a unique name? Does it need
      to be different from a wire that is in a different group? 

- `wire_groups` and `port_groups` can have a project level global unique name
  for identification
    - since they are identified by their name, a `dict` is a more apt structure
      to store the wire and port groups:
        - `get_wire_group(<group_name>)` becomes
          `global_wire_groups['<group_name>']`
        - still iterable like a list, for example with:
            ```python
            for wire_group_name, wire_group in global_wire_groups:
                print(f"{wire_group_name}: {wire_group}")
            ```
- `wire` and `port` in same group need to have diferent names
    - wire and port groups are instantiated in the same verilog context, which
      does not allow multiple ports or wires with the same name
- `wire` and `port` in different group can have same names
    - for example two different verilog modules in the same project can have
      wires with the same name and different widths:
        - module A with a 4 state machine has `wire [1:0] pc`
        - module B with a 6 state machine has `wire [2:0] pc`
    - wires and ports cannot be uniquely identified only by their name at a
      project level

> - We should discuss what datatypes the connect argument should receive. For
  example, if it receives a string, should it assume it is a wire name or a
  wire_group name?
  
 - if even standalone created ports and wires are assigned a global group, can we 
   always search for a wire_group or port_group?

iob_soc.py
arturum1, line 1: In the next PRs, we may want to add the `create_conf`, `create_reg`, and `create_instance` methods, to keep it consistent with the methods of creating wires and ports.

arturum1, line 1: > > * We should discuss how to extract individual wires from groups. For now I have
> >   used the `get_wire_from_group(<group_name>, <wire_name>)` function
> 
> > * We should also discuss if wires and wire_groups should have unique names,
> >   since they are global (accessible by any component in the entire project).
> >   - Does a wire that is in a group need to have a unique name? Does it need
> >   to be different from a wire that is in a different group?
> 
> * `wire_groups` and `port_groups` can have a project level global unique name
>   for identification
>   
>   * since they are identified by their name, a `dict` is a more apt structure
>     to store the wire and port groups:
>     
>     * `get_wire_group(<group_name>)` becomes
>       `global_wire_groups['<group_name>']`
>     * still iterable like a list, for example with:
>       ```python
>       for wire_group_name, wire_group in global_wire_groups:
>           print(f"{wire_group_name}: {wire_group}")
>       ```
> * `wire` and `port` in same group need to have diferent names
>   
>   * wire and port groups are instantiated in the same verilog context, which
>     does not allow multiple ports or wires with the same name
> * `wire` and `port` in different group can have same names
>   
>   * for example two different verilog modules in the same project can have
>     wires with the same name and different widths:
>     
>     * module A with a 4 state machine has `wire [1:0] pc`
>     * module B with a 6 state machine has `wire [2:0] pc`
>   * wires and ports cannot be uniquely identified only by their name at a
>     project level

I have not yet given much thought to the implementation of global wires (and groups) in a Verilog context.

I assume that if we connect a port of an instance to a (global) wire that is in a completely different module, the required ports and wires will be automatically generated in Verilog to make that connection. Probably by using a script that keeps adding ports and wires, climbing up the hierachy tree of Verilog modules, until it finds a common point with the global wire we want to connect to. But I'm not sure if this is what we want.

arturum1: From what I understand, ports are always local to the module that we are instantiating.
While wires are always global for the entire project.

So, multiple instances of the same module can have ports with the same name.
However since [now each port automatically creates a corresponding global wire](https://github.com/IObundle/iob-soc/pull/813#discussion_r1527692870), we need to be careful to ensure that the wire has a different name.
I'm not sure how we should specify the new names for the wires. Should this be done automatically? Should we keep the `connect` argument in the `create_port` and `create_port_group` methods to specify the wire name?

arturum1, line 1: > > * We should discuss what datatypes the connect argument should receive. For
> >   example, if it receives a string, should it assume it is a wire name or a
> >   wire_group name?
> 
> * if even standalone created ports and wires are assigned a global group, can we
>   always search for a wire_group or port_group?

I agree, we could always search for the group name if we pass it a string.
But I think we would still need a way to extract individual wires from groups, like I did here: https://github.com/IObundle/iob-soc/blob/b3765b71e2cfe4540777af2e32e0ee167e2763a9/iob_soc.py#L281

In this example, we don't want to connect the entire 'clk_en_rst' group to the cpu (because the 'arst' signal is different).
So I extracted the 'clk' signal from the group to connect to that port.
I'm not sure what the return value of the `get_wire_from_group()` method is, but I assume it is a reference to a `iob_wire` object.

So, for now, I believe the `connect` argument can receive 2 datatypes:
1) A string that represents a global wire group name;
2) A reference to an `iob_wire` object (that also belongs to a group in the global wire group list).

arturum1: We no longer have a 'global' wire group list. Now it is local.

iob_soc.py
jjts: only groups connect , even if single element groups

iob_soc.py
```diff
@@ -302,19 +277,24 @@ def __init__(self, *args, **kwargs):
...
                 "USE_EXTMEM": USE_EXTMEM
             },
-            io_connections={
-                "clk_i": clk,
-                "arst_i": cpu_reset,
-                "cke_i": cke,
-                "boot_i": boot,
-                "trap_o": trap,
+            connect={
+                "clk": get_wire_from_group("clk_en_rst", "clk"),
```
jjts, line 281: connect the whole group, not individual ports 

ports cannot have any logic or concatenations or part selections

iob_soc.py
arturum1, line 1: When we want to create a port group with a single port, should we use the `create_port` function? Or should we always use the `create_port_group` function?

In this example, I used the `create_port` function to create the `trap` port, and assign it to a (new) group via the `group` argument.
https://github.com/IObundle/iob-soc/blob/327da9172425de1e6f9a7a41fc6ac11a233893f8/iob_soc.py#L192-L198

Should we use something like this instead?
```
        create_port_group(
            name="trap",
            ports=[
                port(
                    name="trap",
                    direction="output",
                    width=1,
                    descr="CPU trap signal",
                )
            ]
        )
```

arturum1: Also, I think we can omit the `group` argument from the `create_port` method. The group name could be equal to the port name by default.

iob_soc.py
arturum1: Removed individual port connections in commit [e789415](https://github.com/IObundle/iob-soc/pull/813/commits/e7894154e966b1e9bd2984d058d37fc785a9da72).
In this commit, I created a dedicated `cpu_clk_en_rst` wire group for the CPU.
In the future, we may use the standard `clk_en_rst` group, but for now, without refactoring the iob_soc.v and iob_picorv32.v modules, we use the dedicated `cpu_clk_en_rst` wire group.

iob_soc.py
arturum1: Should these `create_port` lines be modified?
This is related to this comment: https://github.com/IObundle/iob-soc/pull/813#discussion_r1529434278

iob_soc.py
P-Miranda: If all ports are part of a group and we only work with group connections, I think we should just use the `create_port_group()` function: is more explicit and matches other function using port groups

iob_soc.py
jjts: Port_group should be called simply port, because it is user exposed. The current port should be named port_signal , a more elaborate name but  it is hidden from the user.

arturum1: Should we do the same for wires?
The current `wire_group` becomes just `wire`

jjts: yes

jjts: but w cn also use iob_wire and iob_bus, I thnk it is better

jjts:
merging so I can make any sense of this

jjts:
merging so I can make some  sense of this

PR #816, opened 2024-03-21
Update `iob_aoi.py` and submodules

arturum1:
Check commits for details

### DIFF
--- a/lib/hardware/modules/iob_aoi/iob_aoi.py
+++ b/lib/hardware/modules/iob_aoi/iob_aoi.py
@@ -1,21 +1,132 @@
 import sys
 
 from iob_module import iob_module
-
-from iob_and import iob_and
-from iob_or import iob_or
-from iob_inv import iob_inv
+from iob_wire import get_wire_element, concat_bits
 
 
 class iob_aoi(iob_module):
-    def __init__(self):
-        super().__init__()
+    def __init__(self, *args, **kwargs):
         self.version = "V0.10"
-        self.submodule_list = [
-            iob_and(),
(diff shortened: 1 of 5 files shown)


lib/hardware/modules/iob_aoi/iob_aoi.py
```diff
@@ -1,21 +1,132 @@
...
+
+        self.create_wire(
+            name="and_ab_in",
+            descr="and ab input",
+            elements=[
+                concat_bits(
+                    get_wire_element("inputs", "a"),
+                    get_wire_element("inputs", "b"),
+                )
+            ],
```
arturum1, line 37: Note, this `and_ab_in` wire only has 1 element. That element is a concatenation of bits from two elements of the `inputs` port of `iob_aoi`.
This concatenation matches the current Verilog implementation of `iob_aoi.v`.

If we want to use multiple elements, instead of concatenating bits into a single element, then we cannot use Verilog parameters to configure the number of inputs of the `iob_and` module.
More details in the description of this commit: https://github.com/IObundle/iob-soc/pull/816/commits/fbefaf57640e456da05686cea502b21bc3e73e0a

jjts:
We should distinguish between Python parameters and Verilog parameters.
Python parameters will allow us to do what Verilog can't do, like having a
parameterizable number of IOs.

jjts:
I see no issues now and will merge. This a very simple example, but it is a powerful demonstration of what can be done.
Note that and, or, in and aoi were added only for the purpose of testing this simple example and will be removed later.

PR #821, opened 2024-04-01
Fix py2hw setup process using previously agreed-upon py2hw syntax

arturum1:
Check commits for details.

This PR introduces very few 'py2hw' syntax changes. There are a few small changes in the `iob_aoi.py` file (and submodules), like renaming `iob_module` to `iob_core`, and `insert_verilog()` to `create_snippet()`.

Most of the major changes are in lib scripts that allow the new setup process to work using the 'py2hw' syntax previously agreed-upon.

One significant change is the renaming of the `iob_module.py` file to `iob_core.py`.
There is a new `iob_module.py` file that is not equivalent to the old one.

For review, I think the best approach is to take a look inside the `iob_aoi.py` and the new `iob_core.py` file.
Also try building the `iob_aoi` module with the following steps to see how the new setup process generates the `iob_aoi.v` automatically:
```
make -C lib sim-run CORE=iob_aoi
```


arturum1:
This PR is currently in draft because there is an issue with the instance names. But otherwise, the build dir and the `iob_aoi.v` seem to be well generated.

arturum1:
Py2hw test for `iob_aoi` module runs successfully with:
`make -C lib sim-run CORE=iob_aoi`

lib/scripts/block_gen.py
```diff
@@ -97,14 +98,13 @@ def get_instance_port_connections(instance):
...
             e_singal = port.e_connect.signals[idx]
             comma = (
                 ","
                 if (port_idx < len(instance.ports) - 1) or (idx < len(port.signals) - 1)
                 else ""
             )
-            instance_portmap += (
-                f"        .{signal['name']}({e_singal['name']}){comma}\n"
-            )
+            instance_portmap += f"        .{port_name}({e_singal['name']}){comma}\n"
```
P-Miranda: `signal` instead of `singal`

PR #823, opened 2024-04-08
added yosys-abc and volare,fix syntax error,end synthesys with genus

Vasco-Luz:
Added yosys-abc(probably the standalone yosys will be added in the next pr), added volare to nix(open pdk package manager), fixed an verilog syntax error added by my, a wire had the name do and it a reserved name, changed it to doo. Fixed the way the synthesis looks at SDC files and deactivated latches errors. the reason is because pico32 is ip that cannot be changed, and its the source of the latch. in the future  a standalone synthesis script for pico32 will be made.

### DIFF
--- a/submodules/LIB/scripts/yosys-abc.nix
+++ b/submodules/LIB/scripts/yosys-abc.nix
@@ -1,0 +1,67 @@
+# Copyright 2023 Efabless Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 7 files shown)


jjts:
can anybody explain the .patch files? what applies them?

Vasco-Luz:
Yes basicly the version of openlane2 uses custom instalation of the various tools and it uses localized patch files for instalation. Because of compability versions they only use certain hashes of the various tools. One of the patches is requires for yosys-abs, the other two patches are for yosys.

submodules/LIB/scripts/yosys-abc.nix
```diff
@@ -0,0 +1,67 @@
...
+    owner = "YosysHQ";
+    repo = "abc";
+    rev = rev;
+    sha256 = "sha256-sMBCIV698TIvU/sgTwgPFWDC1kl2TeGv+3pQ06gs7aM=";
+  };
+  yosys_abc = pkgs.clangStdenv.mkDerivation rec {
+    name = "yosys-abc";
+    src = yosys_abc_src;
+    patches = [
+      ./patches/yosys/abc-editline.patch
```
Vasco-Luz, line 49: here is goes looking for the patch

Vasco-Luz:
Ok it makes more sense to push this first to the memwrap.

PR #826, opened 2024-04-09

Vasco-Luz:
Added yosys-abc(probably the standalone yosys will be added in the next pr), added volare to nix(open pdk package manager), fixed an verilog syntax error added by my, a wire had the name do and it a reserved name, changed it to doo. Fixed the way the synthesis looks at SDC files and deactivated latches errors. the reason is because pico32 is ip that cannot be changed, and its the source of the latch. in the future a standalone synthesis script for pico32 will be made.

The patches are because to ensure that only a certain version of yosys is installed for compability problems.


Vasco-Luz:
The ku check error I think is not related to this commit

hardware/common_src/iob_soc_mwrap.v
```diff
@@ -120,7 +120,7 @@ iob_soc #(
     .addr_spram_o(addr),
     .wstrb_spram_o(we),
     .wdata_spram_o(di),
-    .rdata_spram_i(do),
+    .rdata_spram_i(doo),
```
jjts, line 123: change name to spram_do, add the spram_ suffix to all other wires above

Vasco-Luz: OK will do

jjts: I meant prefix

Vasco-Luz: i think is fixed. there was a mistake i made with the wires in the SPRAM that i didnt notice. i think i fixed it.

jjts: you changed the ports, not the wires

jjts: I asked you to change the wires , not the ports

Vasco-Luz: sorry

Vasco-Luz: i added spaces to the ports but will remove them asap. i think the wires related to the SPRAM, now all have SPRAM prefix.

PR #827, opened 2024-04-09

Vasco-Luz:
Added yosys-abc(probably the standalone yosys will be added in the next pr), added volare to nix(open pdk package manager), fixed an verilog syntax error added by my, a wire had the name do and it a reserved name, changed it to doo. Fixed the way the synthesis looks at SDC files and deactivated latches errors. the reason is because pico32 is ip that cannot be changed, and its the source of the latch. in the future a standalone synthesis script for pico32 will be made.

The patches are because to ensure that only a certain version of yosys is installed for compability problems.


Vasco-Luz:
I closed the previous pr in a stupid way when i was reseting the commits.

PR #828, opened 2024-04-09

Vasco-Luz:
Added Yosys (with Yosys-ABC), included Volare into Nix (an open PDK package manager), rectified a Verilog syntax error introduced by myself; a wire named "do" was used, which is a reserved name, so changed it to "doo." Also, modified the synthesis process to properly handle SDC files and deactivated latch errors if pico32.v is present in the sources. 

Yosys is an open-source framework for Verilog RTL synthesis. It takes Verilog code as input and produces a synthesized netlist as output, which can then be used for further downstream processing, such as placement and routing. Yosys supports a wide range of synthesis optimizations and transformations, allowing users to achieve efficient and optimized hardware designs.

Yosys-ABC is an extension of the Yosys synthesis framework, integrating the ABC (A System for Sequential Synthesis and Verification) tool for logic synthesis. ABC is a highly optimized and widely used academic synthesis tool developed at the University of California, Berkeley.

Volare is a package manager designed specifically for managing Process Design Kits (PDKs) in the context of semiconductor manufacturing and integrated circuit (IC) design. It simplifies the installation and management of PDKs, which contain the necessary technology files, libraries, and models for designing integrated circuits.

The patches are requirements for the specific versions of Yosys-abc and Yosys

### DIFF
--- a/submodules/LIB/hardware/syn/build.tcl
+++ b/submodules/LIB/hardware/syn/build.tcl
@@ -24,8 +24,10 @@ set_db init_hdl_search_path $INCLUDE
 set_db stdout_log genus.log
 set_db information_level 7
 set_db super_thread_debug_directory st_part_log
-
-set_db hdl_error_on_latch true 
+set_db hdl_error_on_latch true
+if {[file exists "syn_build.tcl"]} {
+    source "syn_build.tcl"
+}
 set_db lp_power_analysis_effort medium
 set_db lp_power_unit uW
 
@@ -124,10 +126,21 @@ check_design -unresolved
 
 # add optimization constraints
 #----------------------------------------------------------------------
-read_sdc -stop_on_error ./$NODE/$NAME\_dev.sdc
-read_sdc -stop_on_error ./src/$NAME.sdc
-read_sdc -stop_on_error ./src/$NAME\_$CSR_IF.sdc
-read_sdc -stop_on_error ./$NAME\_tool.sdc
+if {[file exists ./$NODE/$NAME\_dev.sdc]} {
+    read_sdc -stop_on_error ./$NODE/$NAME\_dev.sdc
+}
+
+if {[file exists ./src/$NAME.sdc]} {
+    read_sdc -stop_on_error ./src/$NAME.sdc
+}
+
+if {[file exists ./src/$NAME\_$CSR_IF.sdc]} {
+    read_sdc -stop_on_error ./src/$NAME\_$CSR_IF.sdc
+}
+
+if {[file exists ./$NAME\_tool.sdc]} {
+    read_sdc -stop_on_error ./$NAME\_tool.sdc
+}
 
 check_timing_intent 
 
--- a/submodules/LIB/scripts/iob_module.py
+++ b/submodules/LIB/scripts/iob_module.py
@@ -740,12 +740,6 @@ class iob_module:
                             ),
                         )
                         continue
-                elif directory == "hardware/common_src":
-                    shutil.copytree(
-                        os.path.join(os.getcwd(), directory),
-                        os.path.join(cls.build_dir, directory),
-                        dirs_exist_ok=False,
-                    )
                 elif directory == "hardware/fpga":
                     # Skip if board_list is empty
                     if cls.board_list is None:
--- a/hardware/common_src/iob_soc_mwrap.v
+++ b/hardware/common_src/iob_soc_mwrap.v
@@ -42,11 +42,11 @@ wire          [     DATA_W-1:0]    d_rdata;
 //
 
 `ifdef USE_SPRAM
-    wire                       en;
-    wire     [SRAM_ADDR_W-3:0] addr;
-    wire     [DATA_W/8-1:0]    we;
-    wire     [DATA_W-1:0]      di;
-    wire     [DATA_W-1:0]      do;
+    wire                       spram_en;
+    wire     [SRAM_ADDR_W-3:0] spram_addr;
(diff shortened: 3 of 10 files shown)


Vasco-Luz:
I made some stupid mistakes. in the first pr i closed because i did a git reset and force pushed something that was equal to the repository i was trying to make a pull request so it closed.
In the second case i just did a bad pr with a commit that was non existant.

Vasco-Luz:
added yosys with yosys-abc. Changed temprary the time for the pc-emul so it to install yosys sucessefull

submodules/LIB/hardware/syn/build.tcl
```diff
@@ -25,7 +25,7 @@ set_db stdout_log genus.log
 set_db information_level 7
 set_db super_thread_debug_directory st_part_log
 
-set_db hdl_error_on_latch true 
+set_db hdl_error_on_latch false
```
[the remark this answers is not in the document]
Vasco-Luz: > You can't disable this for us! We need this!
> 
> You need another if statement: check if the submodules/PICORV32 directory exists and disable the check.
> 

That is a very good solution, will be implemented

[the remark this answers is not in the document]
Vasco-Luz: Because the setup dir is not an argument passed to the tcl file, he checks all the filenames (without the path) in $VSRC, and if picorv32.v is found, the error on latches is disabled.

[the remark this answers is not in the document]
jjts: not acceptable; the script knows exactly where picorv32.v 


Jut ask the team how to do this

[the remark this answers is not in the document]
Vasco-Luz: OK will fix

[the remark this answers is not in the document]
Vasco-Luz: It checks using relative paths if the submodule PICORV32 exists. If it exists, it will pass a flag so the TCL script will allow latches. An important thing: because the setup directory doesn't exist on the server (after the SSH), I gave it different names. This is because when the Makefile runs in the build directory locally it will check the setup dir because it actually exists, Inside the server the setup dir never exists because only the build dir is copied.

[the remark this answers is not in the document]
jjts: setupdir does not exist for external users ; just check if ../src/picorv32.v exists

jjts:
Suggestion:

build.tcl checks if file syn_build.tcl exists in the same directory and sources it if it exists. Inside syn_build the latch errors disabled.

submodules/LIB/hardware/syn/build.tcl
```diff
@@ -24,8 +24,13 @@ set_db init_hdl_search_path $INCLUDE
 set_db stdout_log genus.log
 set_db information_level 7
 set_db super_thread_debug_directory st_part_log
-
-set_db hdl_error_on_latch true 
+set_db hdl_error_on_latch true
+if {[file exists "../src/picorv32.v"]} {
```
jjts, line 28: no need for this if

submodules/LIB/hardware/syn/syn_build.tcl
```diff
@@ -0,0 +1 @@
+set_db hdl_error_on_latch false
```
jjts, line 1: this file should be in iob-soc/hardware/hardware/syn

Not in lib!

Make sure  it is copied to build dir during setup.

jjts: If the user doesn't want to disable the check, she does not include this content in her syn_build.tcl or doesn't include the file at all.

Vasco-Luz: ok implemented. A iob-soc/hardware/syn folder was created. The folder path was already recognized in iob_module  dir_list , so it automaticly copies the syn_build.tl to the buildir/hardware/syn, without having to do changes in iob_module.py.

submodules/LIB/hardware/syn/build.tcl
```diff
@@ -24,8 +24,10 @@ set_db init_hdl_search_path $INCLUDE
 set_db stdout_log genus.log
 set_db information_level 7
 set_db super_thread_debug_directory st_part_log
-
-set_db hdl_error_on_latch true 
+set_db hdl_error_on_latch true
```
jjts, line 27: is this needed? Isn't true the default?

Vasco-Luz: The original code already had it set as true. I didn't even think about that.

jjts: Ok, but I still don't know the answer

Vasco-Luz: i just tested, the standart the latch error is false. i removed the set from the syn_build.tcl and removed from the set from the build tcl, and it gave no synthesis warnings.

submodules/LIB/scripts/iob_module.py
```diff
@@ -740,12 +740,6 @@ def _copy_srcs(cls, exclude_file_list=[], highest_superclass=None):
                             ),
                         )
                         continue
-                elif directory == "hardware/common_src":
-                    shutil.copytree(
-                        os.path.join(os.getcwd(), directory),
```
Vasco-Luz, line 745: this lines is something that i added in another commit but they serve no purpose, because the directory is copied ether way in the last else. and worst maybe they are causing problems with other cores.

PR #829, opened 2024-04-11
Add support for py2hw dictionaries and py2hw JSON files.

arturum1:
This PR introduces support for python dictionaries with core data, and for JSON formatted files with core data. These two new formats can be used as ways of specifying the core to be generated with the py2hw setup process.
Therefore, currently there are 3 ways of specifying a core to generate:
1) Defining a subclass of `iob_core.py` for the core we are creating, and setting its attributes. [Example here.](https://github.com/arturum1/iob-soc/blob/py2hw/lib/hardware/modules/iob_aoi/iob_aoi.py#L8-L117)
2) Creating a python script with a dictionary of attributes for the new core. [Example here.](https://github.com/arturum1/iob-soc/blob/py2hw/lib/hardware/modules/iob_aoi/iob_aoi.py#L121-L234)
 And [calling the `py2hw(<dict>)` method](https://github.com/arturum1/iob-soc/blob/py2hw/lib/hardware/modules/iob_aoi/iob_aoi.py#L246) of `iob_core.py`.
3) Creating a JSON formated `json/<core_name>.json` file in the core's setup dir. [Example here.](https://github.com/arturum1/iob-soc/blob/py2hw/lib/hardware/modules/iob_aoi/json/iob_aoi.json)
And [calling the `read_py2hw_json(<filepath>)` method](https://github.com/arturum1/iob-soc/blob/py2hw/lib/hardware/modules/iob_aoi/iob_aoi.py#L249) of `iob_core.py`.

Currently, the `iob_aoi` core is ready to be setup using any of the three methods.
To setup the `iob_aoi` core using any method, use one of the following commands:
1) `make -C lib sim-run CORE=iob_aoi`
2) `make -C lib sim-run CORE=iob_aoi SETUP_ARGS='py2hw'`
3) `make -C lib sim-run CORE=iob_aoi SETUP_ARGS='json'`

---

As mentioned in commit 49c224c, when searching for submodules, the setup process searches for `<corename>.py` or `<corename>.json`. The file found first is the one used to generate the module.
As mentioned in that commit, the setup process also assumes the corresponding submodule's directory is the same as the one in which those files are located. So this can cause issues if a user provides multiple files (py or json) in the same directory, because the setup process will assume the setup directory is the same for all of them.
The solution I propose in that commit is to add a new attribute in `iob_core.py` that specifies if the setup process should ignore the setup directory of a module completely. 

Since the `iob_and.py` file still exists, the setup process will probably use that file as a submodule of the `iob_aoi.py` core.
However, you can try deleting the `iob_and.py` file, and run the setup of `iob_aoi` again to see how it still generates the `iob_and` submodule well from the JSON file.
```
rm lib/hardware/modules/iob_and/iob_and.py
make -C lib sim-run CORE=iob_aoi SETUP_ARGS='json'
```
The `SETUP_ARGS='json'` in the code above is optional. If we remove this, the iob_aoi will be generated from the python subclass, but the iob_and will be generated from the JSON file (since we deleted its python).

---

I have also added a new `print_py2hw_attributes()` method in `iob_core.py` that prints all the attributes supported by the `py2hw` interface. These attributes can be passed using any of the 3 formats I specified above.
To print the py2hw attributes of the `iob_aoi` module, use:
`make -C lib print-attr CORE=iob_aoi`
The attributes supported by `iob_aoi` are the same as the ones supported by `iob_core` and most of the lib modules.

The attributes will only change if a subclass introduces new attributes. This will likely happen when we create the `iob_soc.py` script (or `iob_system.py`, or equivalent).

---

I've also updated the `set_default_attributes()` method of `iob_base.py`. This is our standard method to specify new attributes for a class.
With this method, we could already define the attribute, ensuring that we don't override its value. We could also already set a datatype to verify that the value is correct.
With the new update, we can also include a description for the attribute (to be printed by the `print_py2hw_attributes` method).
Now, we can also add a dedicated handler for the attribute, that converts py2hw data (likely encoded in a dictionary) into the internal representation (usually an object).

---

~~I've also added the `export_json_from_dict()` method in `iob_core.py` to generate a JSON file from a given py2hw dictionary.
Essentially converting from format `2` to `3`. (I used this method to generate the `iob_aoi` and `iob_and` JSON files).~~

~~I've also tried to implement the `export_json()` method to covert from the internal object representation into a py2hw dictionary.
Essentially converting from format `1` to `2`.
However, this attempt was unsuccessful because the internal representation uses many different formats, and some of them are generated (like the signals of a port using if_gen) so we might have needed some sort of un-generator (to create the port based on the signals) (inverse of if_gen).
I guess we could also store the original data (before it was generated) somewhere, and therefore recover it when generating a dictionary. But I'm not sure this is worth the effort.~~

Edit: Removed this in commit 1f14a43.

---

In this PR, I maintained support for format `1` because formats `2` and `3` depend on it.
In the internal setup process, when we give a JSON file (format `3`), it converts it to a dictionary (format `2`), and then converts it into a python object of the type `iob_core` for internal representation (format `1`).

---

Check commits for more details


arturum1:
I'm not sure why the lib check failed. I cannot reproduce it in my machine.

P-Miranda:
> I'm not sure why the lib check failed. I cannot reproduce the error in my machine.

inability to reproduce failed CI errors is related with some environment difference:
- try to execute on the same self-hosted runner machine

if the error cannot be reproduced: 
- try to run as `ghactions` user to check for differences in user environment

if the error cannot be reproduced, should be related with the actions runner work directory that left files from previous CI runs:
```bash
# change to ghactions user
su ghactions
# go to self-hosted github actions runner work directory (this example for iob-soc)
cd /home/ghactions/actions-runner/_work/iob-soc
# delete all files except for iob-soc repository
find . -maxdepth 1 -not -name "iob-soc" -exec rm -rf {} +
```

* * *
For this case, I removed the folders in `/home/ghactions/actions-runner/_work/iob-soc` for both self-hosted runners and was able to run the CI job without errors.
So it should be fixed in the next CI job execution.
If the problem comes back, we need to improve the `clean` targets to remove files from previous runs.

PR #833, opened 2024-04-23
Changed from hardcoded to dinamic io inside memwrap.v related to iob-soc instance.

Vasco-Luz:
When the IOB SoC was instantiated inside the MemWrap, its IO ports were hardcoded. This caused issues with other cores that required the IOB SoC ports to be dynamic. A modification was implemented to address this. This pull request contains the necessary changes.

### DIFF
--- a/hardware/common_src/iob_soc_mwrap.v
+++ b/hardware/common_src/iob_soc_mwrap.v
@@ -21,6 +21,84 @@ module iob_soc_mwrap #(
    `include "iob_soc_io.vs"
 );
 
+`ifdef IOB_SOC_USE_EXTMEM
+    wire [      AXI_ID_W-1:0] axi_awid;
+    assign axi_awid_o = axi_awid;
+    wire [    AXI_ADDR_W-1:0] axi_awaddr;
+    assign axi_awaddr_o = axi_awaddr;
+    wire [     AXI_LEN_W-1:0] axi_awlen;
+    assign axi_awlen_o = axi_awlen;
+    wire [             3-1:0] axi_awsize;
+    assign axi_awsize_o = axi_awsize;
+    wire [             2-1:0] axi_awburst;
+    assign axi_awburst_o = axi_awburst;
+    wire [             2-1:0] axi_awlock;
+    assign axi_awlock_o = axi_awlock;
+    wire [             4-1:0] axi_awcache;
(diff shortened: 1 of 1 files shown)


hardware/common_src/iob_soc_mwrap.v
```diff
@@ -21,6 +21,84 @@ module iob_soc_mwrap #(
    `include "iob_soc_io.vs"
 );
 
+`ifdef IOB_SOC_USE_EXTMEM
+    wire [      AXI_ID_W-1:0] axi_awid;
+    assign axi_awid_o = axi_awid;
+    wire [    AXI_ADDR_W-1:0] axi_awaddr;
```
jjts, line 27: where are these wires used ?

In any case, they are not needed : there are snippets that use the _o and _i directly , no assignment needed .
Please ask the team.

Is this module only used for synthesis?

Vasco-Luz: This is a fix so the main can be used in other cores. Just like the iob_memwrap the instance iob_soc needs to use a. Vs for the ios to get the new io, when the iob has alterations, for example in Iob_soc_tester. The wires are used because the .vs of the iob_soc generates ios for the axi that are like .axi_awid_o(axi_awid). Só this is not related to synthesis.

P-Miranda (approved):
`IOB_SOC_USE_EXTMEM` wires are used to connect to `axi_m_portmap` wires in iob_soc instance.

Current `if_gen.py` does not support generating a portmap with sufixes for wires.
Removing the wires and connecting directly to `iob_soc_mwrap.v` ports requires adding option to add sufix to `connection_name` in portmaps:

```python
# if_gen.py module
def m_portmap(port_prefix, wire_prefix, fout, bus_start=0, bus_size=1):
    for i in range(len(table)):
        if table[i]["master"] == 1:
            port = port_prefix + table[i]["name"] + suffix(table[i]["signal"]) # NOTE: port gets suffix
            connection_name = wire_prefix + table[i]["name"] # NOTE: no option to give sufix to connection_name for portmap
            write_portmap(
                port,
                connection_name,
                table[i]["width"],
                bus_start,
                bus_size,
                table[i]["description"],
                fout,
            )
```

AndreMerendeira:
This is a special case of portmap .vs since it mixes internal and external connections, hence the fix

PR #834, opened 2024-04-29
Changed makefile to allow a choice related to the synthesiser used

Vasco-Luz:
Added two dirs for SYN, one for a build.mk to run genus and another to run yosys. To this point yosys dir is empty. In this point for testing the NODE variable is not being passed in the setup Makefile but eventually it will pass. just to be easier to make tests.


submodules/LIB/hardware/syn/Makefile
```diff
@@ -9,6 +9,9 @@ include ../../config_build.mk
 
 REMOTE_BUILD_DIR=$(USER)/$(BUILD_DIR_NAME)
 REMOTE_SYN_DIR=$(REMOTE_BUILD_DIR)/hardware/syn
+PDK_ROOT := $(CURDIR)
+export PDK_ROOT
```
Vasco-Luz, line 13: this need to be added, the reason is because the pdk is installed where PDK_ROOT points, since the location of instalation want to be know, the this need to be exported

submodules/LIB/hardware/syn/Makefile
```diff
@@ -28,6 +31,10 @@ OUTPUT_DIR=results
 #default node
 NODE ?= umc130
 
+#open PDK NODE
+Open_NODE ?= sky130
+Open_hash ?= 6d4d11780c40b20ee63cc98e645307a9bf2b2ab8
```
jjts, line 36: what is this ?

submodules/LIB/hardware/syn/Makefile
jjts: who or what installs the PDK? Is it nix?

If somebody download iob.-soc will the pdk be installed?

submodules/LIB/hardware/syn/Makefile
jjts: is there a tag? a version? why do we need to know a hash??

submodules/LIB/hardware/syn/Makefile
```diff
@@ -42,7 +49,10 @@ UFLAGS+=COV=$(COV)
 UFLAGS+=COV_TEST=$(COV_TEST)
 
 build: $(VHDR) $(VSRC) config.tcl
-ifeq ($(SYN_SERVER),)
+ifeq ($(SYNTHESIZER),yosys)
+	make pdk_install
```
jjts, line 53: so everytime you run synth you need to install the pdk? doesn't make much sense...

jjts: if you cant use nix we need to find a better solution

submodules/LIB/hardware/syn/Makefile
Vasco-Luz: The open node is the pdk to this day exist 2 gf180 and sky130, but the hash it is not required.

submodules/LIB/hardware/syn/Makefile
Vasco-Luz: Volare it's the open source pdk package manager, it's installed in nix. Maybe I can download the pdk to be inside the nix files. Install pdk with volare it basicly checks if the pdk is locally installed, if it's not it will download the pdk, if it's installed it will do nothing.

submodules/LIB/hardware/syn/genus/build.mk
```diff
@@ -0,0 +1,61 @@
+# (c) 2022-Present IObundle, Lda, all rights reserved
+#
+# This makefile is used at build-time
```
jjts, line 3: This makefile segment

jjts (changes requested):

submodules/LIB/hardware/syn/genus/build.mk
Vasco-Luz: leave or remove?

jjts: It is not a makefile. The .mk are makefile **segments**

Vasco-Luz: changed to makefile segment

Vasco-Luz:
I am having a problem with yosys, yosys when synthesising iob_soc.v gives an error related to bsp.vh. i added bsp.vh to the path of .vh,but is not solving the problem.

PR #838, opened 2024-05-05
Almost all hw modules converted to py2hwsw syntax

Edw590:
Some modules don't pass the tests because autogenerated files are missing but the big majority passes.

Modules iob_split, uart and merge were not converted because of being more complex.


jjts:
Why do they need auto-generated files? They have their verilog.

Edw590:
The Verilog files include .vs files. Some of those are not being generated

lib/hardware/modules/div/iob_div_subshift/iob_div_subshift.py
```diff
@@ -1,88 +1,44 @@
...
-        iob_div_subshift()
+def setup(py_params_dict):
+    attributes_dict = {
+        "original_name": "iob_div_subshift",
+        "name": "iob_div_subshift",
+        "version": "0.1",
+        "generate_hw": False,
+        "ports": [
+            {
+                "name": "clk_en_rst",
```
arturum1, line 9:
```suggestion
                "name": "clk_en_rst",
                "type": "slave",
```
This port is missing the `type` key.
Without this key, the scripts assume the interface is of type 'master' and are therefore generating ports in the wrong direction.

arturum1: The `type` key also seems to be missing from the `clk_en_rst` port of other modules in this PR.

arturum1 (changes requested):

lib/hardware/modules/iob_timer/iob_timer.py
arturum1, line 1: To check if the scripts are generating the core correctly, I'm running this command to setup:
`py2hwsw.py iob_timer setup --no_verilog_lint`
Followed by this command to run with Icarus Verilog:
`make -C ../iob_timer_V0.1/ sim-run`

lib/hardware/modules/div/iob_div_subshift/iob_div_subshift.py
arturum1: Resolved in https://github.com/IObundle/iob-soc/pull/840

arturum1 (approved):
New fixes in PR https://github.com/IObundle/iob-soc/pull/840

PR #842, opened 2024-05-08
added synthesiser choice in the makefile and a submodules/LIB/hardware/syn change

Vasco-Luz:
added a choice of synthesiser choise in the makefile. and Made a change in submodules/LIB/hardware/syn so it imports a build.mk in function of the passes synthesiser in the makefile


submodules/LIB/hardware/syn/genus/build.mk
```diff
@@ -0,0 +1,61 @@
...
+	@echo "set VSRC [glob $(VSRC)]" >> $@
+
+clean:
+	find . -maxdepth 1 -type f -not \( -name Makefile -o -name \*.tcl -o -name \*.sdc -o -name syn_build.mk \) -delete
+	@rm -rf ../../*.vh config.tcl fv results st_part_log
+ifneq ($(SYN_SERVER),)
+	ssh $(SYN_SSH_FLAGS) $(SYN_USER)@$(SYN_SERVER) 'if [ -f $(REMOTE_SYN_DIR)/Makefile ]; then make -C $(REMOTE_SYN_DIR) $@; fi'
+endif
+
+.PHONY: run clean config.tcl synth
```
jjts, line 61: Confug.tcl is a file! Not a phony target.

submodules/LIB/hardware/syn/genus/build.mk
```diff
@@ -0,0 +1,61 @@
...
+ifneq ($(wildcard syn_build.mk),)
+include syn_build.mk
+endif
+
+OUTPUT_DIR=results
+
+#default node
+NODE ?= umc130
+
+SYN_SERVER=$(CADENCE_SERVER)
```
jjts, line 20: This part should not be repeated for each tool. See simulation.

jjts: See the FPGA folder, this is wrong . This way you need to repeat these commands for each syn tool.

submodules/LIB/hardware/syn/genus/build.mk
Vasco-Luz: I will change but that was in the original makefile.

submodules/LIB/hardware/syn/genus/build.mk
Vasco-Luz: OK will change, I done it like this because other sythesiser's maybe won't need a cadence server.

jjts: This PR shows it was deleted from the makefile but it should not have been deleted

Why pass that code to the build.mk since it is not specific of the synt tool?


Close this PR and do a new one pls.

submodules/LIB/hardware/syn/genus/build.mk
```diff
@@ -0,0 +1,61 @@
...
+SYN_SYNC_FLAGS=$(CADENCE_SYNC_FLAGS)
+
+synth: $(VHDR) $(VSRC)
+	genus -batch -files genus/build.tcl
+
+
+UFLAGS+=COV=$(COV)
+UFLAGS+=COV_TEST=$(COV_TEST)
+
+build: $(VHDR) $(VSRC) config.tcl
```
jjts, line 33: all this code is unrelated to build.mk for genus

PR #843, opened 2024-05-08
Corrections in Verilog code of hw modules

Edw590:


lib/hardware/modules/iob_nco/hardware/simulation/src/iob_nco_tb.v
```diff
@@ -1,11 +1,16 @@
 `timescale 1ns / 1ps
+`include "iob_reg_conf.vh"
 
-`include "iob_utils.vh"
+`define IOB_CLOCK(CLK, PER) initial CLK=0; always #(PER/2) CLK = ~CLK;
```
jjts, line 4: What happened to iob_utils.vh?

PR #844, opened 2024-05-09
Update Nix to use new `py2hwsw` version

arturum1:
This PR includes changes from https://github.com/IObundle/iob-soc/pull/843. That one should be merged first.

- Update `py2hwsw` commit in Nix, to match the new version in https://github.com/IObundle/py2hwsw/pull/5
  Lib module's tests are now running correctly. Some of them fail due to lib modules that still need to be updated.
  New version of `py2hwsw` no longer generates `__pycache__` directories when running the setup process.
  
Use the following command to launch all lib tests manually:
`make -C lib/ sim-test`
Use the following command to clean all test build directories:
`make -C lib/ sim-clean`


arturum1:
The `lib` test in this PR still gave a permission error. I think this may be due to cached files from previous runs.
I have already deleted those files from the IObundle machines.
Please re-run ghactions in this PR to see if the issue is solved.

PR #847, opened 2024-05-15
feat(fpga): add Kintex-US and Kintex-US+ support

P-Miranda:
- add Kintex-US KU085 and Kintex-US+ KU15P board support


jjts:
do these boards exist or it's a made-up board so we can support de device (FPGA)

For  example, the KU040 is a real board

jjts:
I dont see the need for another kintex us , ku040 already is

PR #848, opened 2024-05-16
Remove macros from some modules

Edw590:
I couldn't remove all macros though. IOB_PULSE and IOB_RESET hang the simulation when I replace with an equivalent task. Haven't figured out yet how to fix that. The other macros were removed and replaced by functions in the new iob_functions module or by the new iob_clock module.

### DIFF
--- a/lib/hardware/modules/iob_clock/iob_clock.py
+++ b/lib/hardware/modules/iob_clock/iob_clock.py
@@ -1,0 +1,38 @@
+def setup(py_params_dict):
+    attributes_dict = {
+        "original_name": "iob_clock",
+        "name": "iob_clock",
+        "version": "0.1",
+        "confs": [
+            {
+                "name": "CLK_PERIOD",
+                "type": "P",
+                "val": "10",
+                "min": "",
+                "max": "",
+                "descr": "Clock period",
+            },
+        ],
+        "ports": [
+            {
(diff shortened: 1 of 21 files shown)


jjts:
why not  use simulation modules ?

lib/hardware/modules/iob_clock/iob_clock.py
```diff
@@ -0,0 +1,38 @@
...
+                "name": "clk",
+                "descr": "Output clock",
+                "signals": [
+                    {"name": "clk", "width": "1", "direction": "output"},
+                ],
+            },
+        ],
+        "snippets": [
+            {
+                "outputs": ["clk_o"],
```
jjts, line 27: what is this for?

Edw590: I needed to declare `clk` as a reg. It seems that's not possible to do on the Python scripts after asking Artur. So instead I declared it on code and assigned the output to it.

arturum1: The snippets 'outputs' list will be used to generate 'global signals', as described here: https://github.com/IObundle/py2hwsw/blob/main/py2hwsw/scripts/iob_core.py#L172

For now, the 'global signals' are not implemented, so that list is not yet used.

PR #849, opened 2024-05-18
added a choice of synthesiser and added yosys synthesis

Vasco-Luz:
added a synthesis step using Yosys and implemented a feature so that when you run make syn-build, you can pass a parameter called SYNTHESIZER to specify whether synthesis should be performed with Yosys or Genus.


jjts:
This PR will be merged and then fixed when I have time

PR #860, opened 2024-07-11
Merge branch 'main' into if_gen2

arturum1:
Some new changes from iob-soc/main, like python scripts, were merged into the py2hwsw repo instead.
Related PR: https://github.com/IObundle/py2hwsw/pull/9.

Lib tests are working. Use this command to test lib modules:
`make -C lib sim-test`


arturum1:
Ghactions 'lib' test failed because 'feynman' machine is using an older nix version and fails to build derivation.
It works in nix version `2.22.0`.

jjts:
Please go ahead and update nix on feynman

jjts:
We should find a way to stick to a version of nix and nix-pkgs in
default.nix, not just import any version.

arturum1:
Fixed Nix issue in commit https://github.com/IObundle/iob-soc/pull/860/commits/e1b98506972acdd8128940cb98a9e5a9a4669320. Ghactions lib test passes.

PR #862, opened 2024-07-17
fix enable

mbot27:
add a new enable condition


jjts:
rejected for training

PR #865, opened 2024-07-18
Generate iob_reg with py2hwsw

mbot27:
deleted iob_reg.v and iob_reg.py automatically generates iob_reg.v


arturum1 (approved):
Changes look good.
I'm not sure why github actions is not starting.

lib/hardware/modules/reg/iob_reg/iob_reg.py
```diff
@@ -23,7 +22,7 @@ def setup(py_params_dict):
             },
             {
                 "name": "RST_POL",
-                "type": "M",
+                "type": "F",
```
jjts, line 25: why change to F ?

jjts: this can be a python variable ; RST_POL applies to the whole design

lib/hardware/modules/reg/iob_reg/iob_reg.py
```diff
@@ -34,16 +33,12 @@ def setup(py_params_dict):
...
                 "signals": [],
             },
             {
                 "name": "io",
                 "type": "master",
-                "port_prefix": "",
-                "wire_prefix": "",
                 "descr": "Input and output",
                 "signals": [
                     {
```
jjts, line 44: separate these signals in different ports. putting them in the same group gives issues 

only create groups if you are sure the signals always connect together ; otherwise separate them

jjts:
lib check failed

lib/hardware/modules/reg/iob_reg/iob_reg.py
arturum1: I also noticed this, but I wasn't sure how we were going to implement it globally in python.
Maybe using `sys.argv`?

jjts: as top level macro ?

jjts: i2s or png should have this

PR #866, opened 2024-07-19
Generate other lib modules 

mbot27:
deleted iob_add2.v, iob_demux.v, iob_int_sqrt.v, iob_iobuf.v, iob_mux.v and they are generated automatically by the iob_add2.py, iob_demux.py, iob_int_sqrt.py, iob_iobuf.py, iob_mux respectively


lib/hardware/modules/iob_demux/iob_demux.py
```diff
@@ -3,7 +3,74 @@ def setup(py_params_dict):
...
+                        "name": "data",
+                        "width": "N*DATA_W",
+                        "direction": "output",
+                    },
+                ],
+            },
+        ],
+        "snippets": [
+            {
+                "outputs": ["data_o"],
```
arturum1, line 61: The 'outputs' list of the snippets were originally intended to reference signals (like the signal `data`) and not groups of signals (like the port `data_o`).
I don't think we can reference entire ports, because some of them may contain input signals as well.

However, trying to reference "data" here would become ambiguous, since we have two signals with the name "data" (in different groups/ports).

lib/hardware/modules/iob_int_sqrt/iob_int_sqrt.py
```diff
@@ -3,7 +3,210 @@ def setup(py_params_dict):
...
+                "name": "rst",
+                "descr": "Input port",
+                "signals": [
+                    {
+                        "name": "rst",
+                        "width": 1,
+                        "direction": "input",
+                    },
+                ],
+            },
```
arturum1, line 78: We can group the clock and reset into a single standard port.
(We don't need to specify the signals because they are generated automatically)
```suggestion
                "name": "clk_rst",
                "descr": "Clock and reset",
                "signals": [],
            },
```

lib/hardware/modules/iob_demux/iob_demux.py
arturum1: I'm not sure how to solve this.
This issue is related to: https://github.com/IObundle/iob-soc/pull/865#discussion_r1683650502

lib/hardware/modules/iob_iobuf/iob_iobuf.py
```diff
@@ -3,7 +3,94 @@ def setup(py_params_dict):
...
+                "name": "o_int",
+                "descr": "o_int wire",
+                "signals": [
+                    {"name": "o_int", "width": 1},
+                ],
+            },
+        ],
+        "snippets": [
+            {
+                "outputs": ["data_o"],
```
arturum1, line 74:
```suggestion
                "outputs": ["o_int", "io", "o"],
```

lib/hardware/modules/iob_iobuf/iob_iobuf.py
```diff
@@ -3,7 +3,94 @@ def setup(py_params_dict):
...
+     `ifdef XILINX
+   IOBUF IOBUF_inst (
+      .I (i_i),
+      .T (t_i),
+      .O (o_int),
+      .IO(io)
+   );
+`else
+   reg o_var;
+   assign io = t_i ? 1'bz : i_i;
```
arturum1, line 85: The generated verilog signal will be `io_io` instead of `io`.
```suggestion
      .IO(io_io)
   );
`else
   reg o_var;
   assign io_io = t_i ? 1'bz : i_i;
```

arturum1 (changes requested):

lib/hardware/modules/iob_mux/iob_mux.py
```diff
@@ -3,7 +3,80 @@ def setup(py_params_dict):
...
+                        "name": "data",
+                        "width": "N*DATA_W",
+                        "direction": "output",
+                    },
+                ],
+            },
+        ],
+        "snippets": [
+            {
+                "outputs": ["data_o"],
```
arturum1, line 61: Same issue as https://github.com/IObundle/iob-soc/pull/866#discussion_r1683739997

lib/hardware/modules/iob_demux/iob_demux.py
jjts: can we reference wire/port.signal ?

jjts:
This PR will not be considered before the previous one is merged

jjts:
I deleted a conversation by mistake 

The outputs of snipets must be signals

It should be possible to disambiguate signals that have the same name by referencing their ports / wires

lib/hardware/modules/iob_demux/iob_demux.py
arturum1: In a previous commit we were planning to be able to specify the wire/port from which the referenced signal came from, using the "wire" attribute:
https://github.com/IObundle/iob-soc/blob/c99f767cb8cc0839a8e96a266384c8a6796697e7/lib/hardware/modules/iob_pulse_gen/iob_pulse_gen.py#L114-L115
In the above example, we are referencing the signal "start_detected_nxt" from the wire "start_detected_nxt".

We deprecated the option to specify the wire in this commit:
https://github.com/IObundle/iob-soc/commit/30ece37874567fb7de350e94905266f55497940a#diff-e6adb0ff5290fcf15b7fbf94f7f370e3cef12f0210079593f3664165092ca2eaL114-R113

We could revert that commit to be able to specify which wire/port we are referencing.

arturum1:
I added a new comment about the wire/port reference here: https://github.com/IObundle/iob-soc/pull/866#discussion_r1684503542

lib/hardware/modules/iob_demux/iob_demux.py
arturum1: > can we reference wire/port.signal ?

We could do that as well, by updating the py2hwsw scripts to support that syntax `<wire/port name>.<signal name>`.
The `iob_demux.py` snippet we become:
```suggestion
                "outputs": ["data_o.data"],
```
Port name = "data_o"
Signal name = "data"

jjts: port name does not have suffix

arturum1: In the current PR state it does have a suffix. See [here](https://github.com/IObundle/iob-soc/pull/866/files/ad43988ac54cf2f27d83ba6e6e025c668afbe91f#diff-039f692536a7b07fa298628979e24db9b582be4aae8ed2f4a74befca60070254R48).
Maybe that suffix should be removed.

jjts:
Not enough bandwidth. Closing before iob_reg is merged

PR #869, opened 2024-07-24
feat(fifo2axis): improve tlast logic

P-Miranda:
- tlast automatically resets after `len` fifo reads
- this allows to sinal new frames without resetting the module
- modcnt starts at max value, so that after first enable, counter goes to 0 and aligns with frame data


jjts:
This branch is frozen. Please target branch ifgen2 and keep a copy in your working repo .

PR #870, opened 2024-07-24
new iob_reg.py and iob_pulse_gen.py

mbot27:
automatically generates iob_reg.v and iob_pulse_gen.v


jjts:
you need to merge with the current branch before attempting to PR

jjts:
This solution is not what we agreed. RST_POL must be out! Rewrite wart to use iob_reg instead of naked registers.

PR #875, opened 2024-07-26
sync module generated automatically with py2hwsw

mbot27:


lib/hardware/modules/iob_add2/iob_add2.py
```diff
@@ -3,7 +3,79 @@ def setup(py_params_dict):
...
+                "signals": [
+                    {"name": "sum_int", "width": "W+1"},
+                ],
+            },
+        ],
+        "snippets": [
+            {
+                "verilog_code": f"""
+    assign sum_o   = in1_i + in2_i;
+    assign carry_o = sum_o[W];
```
arturum1, line 75: It looks like the [original implementation](https://github.com/IObundle/iob-soc/pull/875/files#diff-c0667e001eea98d153a3d0c2c187ae68f16a235dc89e4bffd43f897cfb353fa7L11-L13) of this module was wrong.
We should use the internal `sum_int` wire to store the result of the sum, including the carry bit. 
```suggestion
    assign sum_int = in1_i + in2_i;
    assign sum_o = sum_int[W-1:0];
    assign carry_o = sum_int[W];
```

lib/hardware/modules/iob_mux/iob_mux.py
```diff
@@ -51,6 +56,23 @@ def setup(py_params_dict):
...
+            assign data_o = data_o_reg;
+            integer input_sel;
+            always @* begin
+                data_o_reg = {DATA_W{1'b0}};
+                for (input_sel = 0; input_sel < N; input_sel = input_sel + 1) begin : gen_mux
+                    if (input_sel == sel_i) begin
+                         data_o_reg = data_i[input_sel*DATA_W+:DATA_W];
+                    end
+                end
+            end
```
arturum1, line 72: The new py2hwsw version 0.7.5, automatically detects output ports of type "reg". So there is no need to create a new internal `data_o_reg` for this module.
```suggestion
            integer input_sel;
            always @* begin
                data_o = {DATA_W{1'b0}};
                for (input_sel = 0; input_sel < N; input_sel = input_sel + 1) begin : gen_mux
                    if (input_sel == sel_i) begin
                         data_o = data_i[input_sel*DATA_W+:DATA_W];
                    end
                end
            end    
```

lib/hardware/modules/sync/iob_reset_sync/iob_reset_sync.py
arturum1, line 1: Since this module was converted to python. We can delete the [iob_reset_sync.v](https://github.com/mbot27/iob-soc/blob/if_gen2/lib/hardware/modules/sync/iob_reset_sync/hardware/src/iob_reset_sync.v) module.

arturum1 (changes requested):

arturum1 (approved):

PR #894, opened 2024-08-08
Move iob-picorv32 to lib and remove submodule

Edw590:


jjts:
the cache test is not supposed to fail: it is sensitive to the CPU location

PR #901, opened 2024-08-11
CPU jumps between programs instead of restarting

Edw590:
CPU starts at address 0x40000000 now, with P_BIT being used in the instruction bus too.

Also merge with if_gen2.


jjts:
Strange start address

Edw590:
It's the address with currently the P_BIT as the bit 30. When it's changed to 31 in the next commits, the start address will be moved to `0x80000000`.

PR #906, opened 2024-08-14
Enable ku040 wrapper; Update py2hwsw version; Remove unused wrapper scripts. 

arturum1:


arturum1:
Ghactions ku040 test failed because fpga is not working.
`ERROR: [Labtoolstcl 44-199] No matching targets found on connected servers: localhost`

arturum1:
It didn't work. I tried programming it locally and it still failed with the same error.
I think we might need to power cycle the fpga board itself.

PR #914, opened 2024-08-20
Update Py2HWSW and remove div modules

agrevin:
This PR removes the div modules generated with Py2HWSW that are now integrated into Py2HWSW.
The div modules in the new version are the same as IObundle/iob-soc#911.
`iob_div_subshift_frac` uses the new fsm attribute instead of a comb.
Running `py2hwsw iob_div_subshift_frac setup`, within the nix-shell environment, raises the following warnings from verilator(sudo privileges might be required):
`%Warning-WIDTHEXPAND: ../iob_div_subshift_frac_V0.1/hardware/src/iob_div_subshift_frac.v:131:25: Operator GTE expects 9 bits on the RHS, but RHS's VARREF 'divisor_i' generates 8 bits.
                                                                                               : ... In instance iob_div_subshift_frac_tb.uut
  131 |         if (res_acc_nxt >= divisor_i) begin
      |                         ^~
                      ... For warning description see https://verilator.org/warn/WIDTHEXPAND?v=5.012
                      ... Use "/* verilator lint_off WIDTHEXPAND */" and lint_on around source to disable this message.
%Warning-WIDTHEXPAND: ../iob_div_subshift_frac_V0.1/hardware/src/iob_div_subshift_frac.v:139:20: Operator ASSIGN expects 2 bits on the Assign RHS, but Assign RHS's CONST '1'h1' generates 1 bits.
                                                                                               : ... In instance iob_div_subshift_frac_tb.uut
  139 |             pc_nxt = 1'b1;
      |                    ^
%Warning-UNUSEDSIGNAL: ../iob_div_subshift_frac_V0.1/hardware/src/iob_div_subshift_frac.v:25:23: Signal is not used: 'divisor_reg'
                                                                                               : ... In instance iob_div_subshift_frac_tb.uut
   25 |     wire [DATA_W-1:0] divisor_reg;
      |                       ^~~~~~~~~~~
`


jjts:
these modules cannot be moved to py2 before they are running

PR #915, opened 2024-08-20
feat(nco): nco improvements

P-Miranda:
- update nco with two clock domains:
    - clk/system clock domain for CSR
    - clk_in source clock for generated clock
- add iob_nco_sync unit


lib/hardware/clocks_resets/iob_nco/hardware/simulation/src/iob_nco_tb.v
```diff
@@ -87,7 +87,8 @@ module iob_nco_tb;
       .iob_rdata_o(iob_rdata_o),
       .iob_ready_o(iob_ready_o),
       .iob_rvalid_o(iob_rvalid_o),
-      .clk_o(clk_out)
```
jjts, line 90: isnt the above an iob_bus ? why explicit portmap ?

lib/hardware/clocks_resets/iob_nco_sync/iob_nco_sync.py
```diff
@@ -0,0 +1,205 @@
+def setup(py_params_dict):
```
jjts, line 1: this module is part of the nco , not a clock or reset

lib/hardware/clocks_resets/iob_nco/hardware/simulation/src/iob_nco_tb.v
P-Miranda: probably due to historical reasons, the testbench was already written this way.
Addressed in latest commit

lib/hardware/clocks_resets/iob_nco_sync/iob_nco_sync.py
P-Miranda: Moved to iob_nco/hardware/modules in latest commit

PR #919, opened 2024-08-21
Use new py2 version; iob_div_subshift_frac now uses fsm attribute

agrevin:


jjts:
this PR is not running tests, why ?

PR #927, opened 2024-08-24
Ram

mbot27:


agrevin:
In the new version of Py2HWSW, the `combs` and `fsms` attributes are now called `comb` and `fsm` respectively.
They are also no longer lists. This means that every instance of comb or list must be updated.
Before:
`"combs/fsms": [
...
],`
After:
`"comb/fsm":
...`

jjts:
go back and change one ram module only

PR #939, opened 2024-08-30
module iob_ram_sp_generated with py2hwsw

mbot27:


lib/hardware/memories/ram/iob_ram_sp/hardware/src/iob_ram_sp.v
```diff
@@ -1,31 +0,0 @@
...
-) (
-   input                     clk_i,
-   input                     en_i,
-   input                     we_i,
-   input      [(ADDR_W-1):0] addr_i,
-   output reg [(DATA_W-1):0] d_o,
-   input      [(DATA_W-1):0] d_i
-);
-
-   //this allows ISE 14.7 to work; do not remove
```
arturum1, line 16: This comment mentions that the localparam is needed for ISE 14.7 to work.
I assume the new false-parameter also works, but I'm not sure.

arturum1 (approved):

PR #951, opened 2024-09-20
Add needed suffixes to accomodate new Py2HWSW syntax

agrevin:
Also removed accumulators, that were moved to py2, and updated py2 version


agrevin:
This PR should allow the tests in IObundle/py2hwsw#47 to pass.

PR #957, opened 2024-10-08
fix(xsim): variables for remote simulation

P-Miranda:
- add simulation variables for remote simulation
- use same variables for VIVADO suite


jjts:
this PR can oly be accepeted after the same change is made in py2

P-Miranda:
> this PR can oly be accepeted after the same change is made in py2

submitted PR in iobundle/py2hwsw#52

PR #960, opened 2024-10-10
feat(reset): improve reset sync

P-Miranda:
- reset sync can accept input with different polarity from output reset
- create iob_r_rstn to use different arst polarity from the rest of the system
- remove logic from fpga wrapper arst with ext mem


jjts:
Not in scope

jjts:
This PR is deprecated and no longer needed

PR #989, opened 2025-02-19
[axis2axi_in] added missing ext_mem_clk_o

AndreMerendeira:


jjts:
This module should not exist any more

jjts:
All modules should be called iob_ something

Issue #1014, opened 2025-05-05
include example build dir


jjts:
For those who prefer to see the code and use their tools, we should supply an example build directory with all files

PR #1015, opened 2025-05-06
feat(gh): release artifacts

P-Miranda:
- initial atempt to generate multiple build directories and upload them as build artifacts for repository releases
- addresses #1014


jjts:
wil this run automatically in each release ?

please explain each artifact. why INITMEM appears twice?

